PR #163, opened 2015-01-05
Hunt fix

parkouss:
implement the ability to search for a bug fix instead of a regression.

### DIFF
--- a/mozregression/bisector.py
+++ b/mozregression/bisector.py
@@ -27,7 +27,8 @@ class BisectorHandler(object):
     """
     build_type = 'unknown'
 
-    def __init__(self):
+    def __init__(self, find_fix=False):
+        self.find_fix = find_fix
         self.found_repo = None
         self.build_data = None
         self.good_revision = None
@@ -58,6 +59,9 @@ class BisectorHandler(object):
         """
         raise NotImplementedError
 
+    def _reverse_if_find_fix(self, var1, var2):
+        return (var1, var2) if not self.find_fix else (var2, var1)
+
     def initialize(self):
         """
         Initialize some data at the beginning of each step of a bisection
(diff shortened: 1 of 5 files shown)


mozregression/main.py
```diff
@@ -79,6 +79,9 @@ def parse_args(argv=None):
                         help=("last known good revision (use with"
                               " --inbound)."))
 
+    parser.add_argument("--hunt-fix", action="store_true",
```
wlach: General comment: I think I slightly prefer the nomenclature of "find fix" as opposed to "hunt fix", just because the former ("find") is more commonly used for these sorts of things. Could you do a global s/hunt/find/ on this? :)

mozregression/bisector.py
```diff
@@ -58,6 +59,9 @@ def _print_progress(self, new_data):
         """
         raise NotImplementedError
 
+    def _reverse_if_hunt_fix(self, v1, v2):
```
wlach: Since we're using this for more than just versions, let's call the variables str1 and str2

mozregression/bisector.py
```diff
@@ -66,20 +70,26 @@ def initialize(self):
         This will only be called if there is some build data.
         """
         self.found_repo = self.build_data[0]['repository']
-        self.good_revision = self.build_data[0]['changeset']
-        self.bad_revision = self.build_data[-1]['changeset']
+        first, last = self._reverse_if_hunt_fix(0, -1)
+        self.good_revision = self.build_data[first]['changeset']
+        self.bad_revision = self.build_data[last]['changeset']
```
wlach: Couldn't you squeeze this all into one line by passing `self.build_data[0]['changeset']` and `self.build_data[-1]['changeset']` as arguments to "reverse_if_hunt_fix"?

mozregression/bisector.py
```diff
@@ -121,8 +132,9 @@ class NightlyHandler(BisectorHandler):
     def initialize(self):
         BisectorHandler.initialize(self)
         # register dates
-        self.good_date = self.build_data.get_associated_data(0)
-        self.bad_date = self.build_data.get_associated_data(-1)
+        first, last = self._reverse_if_hunt_fix(0, -1)
+        self.good_date = self.build_data.get_associated_data(first)
+        self.bad_date = self.build_data.get_associated_data(last)
```
wlach: Likewise, I think this could all be one line?

mozregression/bisector.py
```diff
@@ -255,16 +288,20 @@ def bisect_nightlies(self, good_date, bad_date):
...
                 if days > 7:
                     self._logger.info("At least one build folder was"
                                       " invalid, we have to start from"
                                       " %d days ago." % days)
-                return self.bisect_inbound(infos['changeset'],
-                                           handler.bad_revision)
+                good_rev, bad_rev = infos['changeset'], handler.bad_revision
+                if handler.hunt_fix:
+                    good_rev = handler.good_revision
+                    bad_rev = infos['changeset']
```
wlach: May as well squeeze these two statements into good_rev, bad_rev = handle.good_revision, infos['changeset']. Also, I would put "good_rev, bad_rev = infos['changeset'], handler.bad_revision" in an else clause.

PR #168, opened 2015-01-10
ensure in nightly builds folders that txt info file contains enough info

parkouss:


parkouss:
This is no more relevant and does not address the bug really - I will replace it with another PR.

PR #169, opened 2015-01-10
updating gh-pages and upgrading with jekyll

parkouss:


contribute.md
```diff
@@ -0,0 +1,68 @@
+---
+layout: default
+title: Contribute
+order: 40
+---
+
+# Contribute to mozregression
+
+Contributors are welcome on mozregression!
```
wlach: I think I'd prefer to keep this information on the wiki, in keeping with how we handle other projects in auto-tools. Maybe just use the existing verbiage on mozilla.github.io/mozregression:

See the mozregression project page for more information on how to contribute.

install.md
```diff
@@ -0,0 +1,28 @@
...
+
+# Install or upgrade
+
+mozregression is a Python (currently 2.7) package installable via pip. You can
+follow the OS-specific instructions to install pip if you don't know how to do it:
+
+- **Windows**:
+  Download and install python 2.7 from
+  [ActiveState](http://www.activestate.com/activepython/downloads).
+  It will include pip and set python in your PATH.
```
wlach: Do you think this is a better approach than the previously-suggested use of mozilla-build?

quickstart.md
```diff
@@ -0,0 +1,41 @@
...
+Then mozregression will go on and do the same things with inbound builds
+to get you the smallest possible window range for inbounds.
+
+There is plenty of options for mozregression. See
+
+    mozregression -h
+
+To list them all. Have a look at the
+[usage]({{ "/usage.html" | prepend: site.baseurl }})
+section to go further in mozregression usage!
```
wlach: to go further in mozregression usage! -> to learn more!

install.md
parkouss: I assume it is simpler for users, because that includes pip and set python in the PATH. Your point is good though, because mozilla-build does not just install python - it also install everything needed to build mozilla projects - but do we need this now that mozcommitbuilder has been removed ?
Maybe we could keep the mozilla-build step in case we want to build things in a mozregression future, instead of just downloading binaries.

Honestly, I am not a windows user, and i don't really mind one of the two options - still I would prefer a pure python approach personnaly, but maybe just because I know that the activestate python is easy to install and works well. Your choice !

index.md
```diff
@@ -0,0 +1,24 @@
...
+
+mozregression is an interactive regression range finder for Mozilla *nightly*
+and *inbound* builds. It uses a binary search algorithm for quickly determining
+a changeset range corresponding to when a problem was introduced. It is widely
+used by Mozilla developers and other community members to help find regressions.
+
+Currently, mozregression can works with:
+
+{% comment %}I am not able to write it with line breaks and keep good list formatting{% endcomment %}
+{% for app in site.data.mozregression_supported_apps %}{% assign first = site.data.mozregression_supported_apps | first %}
```
wlach: Would it be possible to link to the relevant projects? People might not know what B2G is. For B2G, please add a paranthetical comment indicating that it refers to FirefoxOS, and only works with the desktop version.

documentation/index.md
```diff
@@ -0,0 +1,17 @@
+---
+layout: documentation
+title: Documentation
+order: 20
+main_link: True
+---
+
+# Documentation
+
+Here you can find some help to use **mozregression**.
```
wlach: We probably don't need this line. The links say it all. :)

documentation/usage.md
```diff
@@ -0,0 +1,90 @@
...
+
+       mozregression --command 'test-command {binary}'
+
+## Network persistence
+
+- Use a folder to keep downloaded files
+
+        mozregression --persist temp
+
+- Use a cache for http transfer
```
wlach: Maybe add a parenthetical "(speeds up downloading the same builds a second time)"

documentation/automatic-bisection.md
```diff
@@ -0,0 +1,96 @@
...
+layout: documentation
+title: Automatic bisection
+doc_order: 10
+description: How to evaluate tested build automatically with a command line
+---
+
+# Automatic bisection
+
+Mozregression allows to evaluate builds automatically with a command line
+instead of interactively asking you.
```
wlach: "allows to" isn't the best grammar, I'd probably say something like:

"As an alternative to using mozregression's interactive mode, it is possible to evaluate builds automatically by writing a custom script. This can save a lot of time if the reproduction steps are complicated or you have a large regression range."

documentation/automatic-bisection.md
```diff
@@ -0,0 +1,96 @@
...
+
+Let's see how that works !
```
wlach: Up to you but I don't think this line is necessary. :)

contribute.md
wlach: I'd probably say something like "We welcome contributions to mozregression!"

contribute.md
```diff
@@ -0,0 +1,15 @@
...
+
+See the [mozregression project page](https://wiki.mozilla.org/Auto-tools/Projects/Mozregression)
+for more information on how to contribute.
+
+Code source is maintained on [github]({{ site.github_url }}).
```
wlach: s/Code source/Source code/

PR #177, opened 2015-01-30
Pregenerate a cache of build information and have mozregression use it

kapy2010:


mozregression/prefetch.py
```diff
@@ -0,0 +1,461 @@
...
+                    result['repository'] = build_info_text['repository']
+                    inbound_info.append(result)
+
+        return inbound_info
+
+    def build_json_files(self):
+        j = json.dumps(self.get_inbound_info(), indent=2)
+        f = open('inbound_%s.json' % self.app_name, 'w')
+        print >> f, j
+        f.close
```
wlach: This line doesn't do anything, you have to call f.close().

I'd probably actually use the with statement here for more conciseness, something like:

with open('inbound_%s.json' % self.app_name, 'w') as f:
   f.write(json.dumps(self.get_inbound_info())

parkouss:
Well, I feel sorry about that but it seems that we are not going this way now - see https://bugzilla.mozilla.org/show_bug.cgi?id=1132151 instead. @kapy2010, thanks for the effort here! We can discuss more about the s3 migration if you'd like to work on this (though yet this requires some work from taskcluster - see url above)!
So, I'm closing this PR. :(

PR #178, opened 2015-01-30
"pep8 check for travis ci build on mozregression"

paked:
Hi guys,
(This is my first bug)

Just finished working on the [pep8 check for travis ci build on mozregression](https://bugzilla.mozilla.org/show_bug.cgi?id=1127575) feature. While making the code PEP-8 compliant I was not exactly focussing on the neatest solution, but more the quickest one, If considered a good idea I will open an issue about neatening up the whole project (if there is not already one).

### DIFF
--- a/mozregression/bisector.py
+++ b/mozregression/bisector.py
@@ -17,6 +17,7 @@ def compute_steps_left(steps):
         return 0
     return math.trunc(math.log(steps, 2))
 
+
 class BisectorHandler(object):
     """
     React to events of a :class:`Bisector`. This is intended to be subclassed.
@@ -127,6 +128,7 @@ class BisectorHandler(object):
     def user_exit(self, mid):
         pass
 
+
 class NightlyHandler(BisectorHandler):
     build_data_class = NightlyBuildData
     build_type = 'nightly'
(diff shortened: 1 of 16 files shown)


mozregression/bisector.py
```diff
@@ -360,8 +365,8 @@ def bisect_nightlies(self, good_date, bad_date):
                     # so we can't go on inbound. Anyway, these are probably too
                     # old and won't even exists on inbound.
                     self._logger.warning("Not enough changeset information to"
-                                         " produce initial inbound regression"
-                                         " range. Builds are probably too old.")
+                                         "produce initial inbound regression"
+                                         "range. Builds are probably too old.")
```
parkouss: Nits: you removed some useful spaces here

mozregression/main.py
```diff
@@ -124,9 +124,9 @@ def parse_args(argv=None):
 
     parser.add_argument("-c", "--command",
                         help=("Test command to evaluate builds automatically."
-                              " A return code of 0 will evaluate build as good,"
-                              " any other value will evaluate the build as"
-                              " bad."))
+                              " A return code of 0 will evaluate build as "
+                              "good, any other value will evaluate the build"
```
parkouss: Could you put the space before good, instead of after the as ? This will be consistent with the rest of the code.

PR #182, opened 2015-02-12
use the package_name info from mozversion to start fennec

parkouss:
Do not merge yet - waiting for mozversion release


parkouss:
ok, mozversion 1.2 is released, this is ok to merge this in now.

PR #184, opened 2015-02-13
Added inbound_prefetch.py that prefetch inbound build info

kapy2010:


parkouss:
Erf, same comment as for #177.

PR #185, opened 2015-02-13
Fix Bug 1128601 - Retry on http get errors in mozregression

imjalpreet:


mozregression/build_data.py
```diff
@@ -257,7 +257,7 @@ def find_build_info_txt(self, url):
         is found.
         """
         data = {}
-        response = get_http_session().get(url)
+        response = retry_get(args=(url))
```
parkouss: I think you're missing a comma here: (url,)

mozregression/utils.py
```diff
@@ -162,7 +172,7 @@ def url_links(url, regex=None, auth=None):
     """
     Returns a list of links that can be found on a given web page.
     """
-    response = get_http_session().get(url, auth=auth)
+    response = retry_get(args=(url, auth))
```
parkouss: This is not working - I suppose you need something like retry_get(args=(url,), kwargs=dict(auth=auth)))

mozregression/utils.py
```diff
@@ -60,6 +61,15 @@ def names(self):
 CACHE_SESSION = None
 
 
+def retry_get(**kwargs):
```
parkouss: Maybe a better way to facilitate calls is to mimic the get behaviour only. (Sorry, I cahnged my maind about it :()
Something like the following will simplify the calls:
def retry_get(url, **kwargs):
    return redo.retry(url, kwargs=kwargs, ...)

imjalpreet: Can you explain what is this **url**?

parkouss: Yes, that's the main argument for the get method. for example when you do:
retry_get(args=(chset_url,))
this is the "chset_ur". The idea is to be able to simply call:
retry_get(chset_url)

imjalpreet: So, basically I have to return redo.retry(get_http_session().get, attempts=3, sleeptime=1, retry_exceptions=(..., ...), args=(url), kwargs=kwargs), right?

parkouss: Yep, exactly. Just don't miss the comma in args=(url,). :)

mozregression/utils.py
```diff
@@ -60,6 +61,14 @@ def names(self):
 CACHE_SESSION = None
 
 
+def retry_get(args, **kwargs):
```
parkouss: Hm I may have been unclear here. It's all good except that I would like the function to be like this:
def retry_get(url, **kwargs):
    return redo.retry(..., args=(url,), kwargs=kwargs)
So a caller will just issue:
retry_get(self.pushlog_url()) instead of retry_get(args=(self.pushlog_url(),))
and
retry_get(url, auth=auth) instead of retry_get(args=(url,), kwargs=dict(auth=auth))

parkouss: Also, check the indentation - It seems to be more than 4 spaces

PR #186, opened 2015-02-15
Bug 1111310 - Allow to configure mozregression with config file(s)

jpigree:
mozregression try to load a configuration file located on
"~/mozregression.cfg" and takes its values as default values
(overridable by command line passed args).

### DIFF
--- a/mozregression/main.py
+++ b/mozregression/main.py
@@ -12,7 +12,10 @@ import mozinfo
 import datetime
 import sys
 import atexit
+import os
+
 from argparse import ArgumentParser
+from ConfigParser import SafeConfigParser, Error
 from mozlog.structured import commandline, get_default_logger
 from requests.exceptions import RequestException
 
@@ -27,6 +30,28 @@ from mozregression.bisector import BisectRunner
 from mozregression.launchers import REGISTRY as APP_REGISTRY
 from mozregression.test_runner import ManualTestRunner, CommandTestRunner
 
+DEFAULT_CONF_FNAME = os.path.expanduser(os.path.join("~",
+                                                     ".mozregression.cfg"))
+
+
+def get_defaults(conf_name):
+    """
+    Get custom defaults from configuration file in argument
+    """
+    defaults = {}
+
+    if os.path.isfile(conf_name):
+        try:
+            config = SafeConfigParser()
+            config.read([conf_name])
+            defaults = dict(config.items("Defaults"))
+            print("%s loaded" % conf_name)
+        except Error as err:
+            sys.exit("Error while parsing %s =>  no custom default values\n%s"
+                     % (conf_name, str(err)))
+
+    return defaults
+
 
 def parse_args(argv=None):
     """
@@ -40,6 +65,7 @@ def parse_args(argv=None):
              " %(prog)s [OPTIONS]"
              " --inbound --bad-rev BAD_REV --good-rev GOOD_REV")
 
+    defaults = get_defaults(DEFAULT_CONF_FNAME)
     parser = ArgumentParser(usage=usage)
     parser.add_argument("--version", action="version", version=__version__,
                         help=("print the mozregression version number and"
@@ -94,6 +120,7 @@ def parse_args(argv=None):
                         help="addon to install; repeat for multiple addons.")
 
     parser.add_argument("-p", "--profile",
+                        default=defaults.get("profile"),
                         metavar="PATH",
                         help="profile to use with nightlies.")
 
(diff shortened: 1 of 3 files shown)


mozregression/main.py
```diff
@@ -11,7 +11,10 @@
 import mozinfo
 import datetime
 import sys
+import os.path
```
parkouss: I think just "import os" is more common here

mozregression/main.py
```diff
@@ -26,6 +29,30 @@
...
+    """
+    defaults = {}
+
+    if os.path.isfile(conf_name):
+        try:
+            config = SafeConfigParser()
+            config.read([conf_name])
+            defaults = dict(config.items("Defaults"))
+            print("%s loaded" % conf_name)
+        except (SystemError, Error) as err:
```
parkouss: Why SystemError ? I understand Error (ConfigParser.Error), but not SystemError here.

mozregression/main.py
```diff
@@ -26,6 +29,30 @@
...
+            print("Error while parsing %s =>  no custom default values\n%s"
+                  % (conf_name, str(err)))
+    else:
+        print("%s not found => no custom default values" % conf_name)
```
parkouss: I don't think this message is useful - don't print anything when no config is read.

mozregression/main.py
```diff
@@ -26,6 +29,30 @@
...
+    defaults = {}
+
+    if os.path.isfile(conf_name):
+        try:
+            config = SafeConfigParser()
+            config.read([conf_name])
+            defaults = dict(config.items("Defaults"))
+            print("%s loaded" % conf_name)
+        except (SystemError, Error) as err:
+            print("Error while parsing %s =>  no custom default values\n%s"
```
parkouss: I think it could be useful to exit here if the file is not well formed.

tests/unit/test_main.py
```diff
@@ -45,6 +46,32 @@ def test_get_usage(self, stdout):
         self.assertEqual(exitcode, 0)
         self.assertIn('usage:', output)
 
+    def test_get_erronous_cfg_defaults(self):
+        temp_file = tempfile.NamedTemporaryFile()
```
parkouss: I suspect that the use of tempfile.NamedTemporaryFile() is the cquse of the error on windows, becquse the file is open in read and write mode. After writing, you may have to to a seek(0) to rewind the file.
I suggest that you use instead:
conf = tempfile.mkstemp()
self.addCleanup(os.unlink, conf) # ensure file will be delete
with open(conf, 'w') as f:
    f.write(...) # write the conf
# now, execute the test on conf

mozregression/main.py
jpigree: Okay. I tried to reproduce the case when the safeConfigParser throws a SystemError but I can't. I removed it.

jpigree:
Yeah. At long last. I have vanquished those windows specificities. :)

tests/unit/test_main.py
```diff
@@ -45,6 +47,46 @@ def test_get_usage(self, stdout):
...
+    def test_get_erronous_cfg_defaults(self):
+        conf = tempfile.mkstemp()
+        handle = conf[0]
+        filepath = conf[1]
+        self.addCleanup(os.unlink, filepath)
+
+        with os.fdopen(handle, 'w') as conf_file:
+            conf_file.write('aaaaaaaaaaa [Defaults]\n')
+
+        try:
```
parkouss: Could you rewrite this with self.assertRaise() instead ? I find it much more readable:
with self.assertRaises(SystemExit):
   main.get_defaults()

tests/unit/test_main.py
```diff
@@ -45,6 +47,46 @@ def test_get_usage(self, stdout):
         self.assertEqual(exitcode, 0)
         self.assertIn('usage:', output)
 
+    def test_get_erronous_cfg_defaults(self):
+        conf = tempfile.mkstemp()
```
parkouss: could you rewrite this as handle, conf = tempfile.mkstemp() ?

tests/unit/test_main.py
```diff
@@ -45,6 +47,46 @@ def test_get_usage(self, stdout):
...
+        valid_values = {'profile': 'PROFILE',
+                        'http-timeout': '10.2',
+                        'app': 'fennec',
+                        'persists': 'true',
+                        'http-cache-dir': '/home/jp/.mozregression',
+                        'repo': 'mozilla-aurora',
+                        'persist': '~/mozpersist',
+                        'bits': '64'}
+
+        conf = tempfile.mkstemp()
```
parkouss: same here: handle, conf = tempfile.mkstemp()

PR #190, opened 2015-02-20
Fix Bug 1132867 - Allow using mozregression with a particular pref

imjalpreet:
Review: @parkouss

### DIFF
--- a/mozregression/main.py
+++ b/mozregression/main.py
@@ -11,6 +11,8 @@ Entry point for the mozregression command line.
 import mozinfo
 import datetime
 import sys
+
+import mozprofile
 from argparse import ArgumentParser
 from mozlog.structured import commandline, get_default_logger
 from requests.exceptions import RequestException
@@ -143,6 +145,15 @@ def parse_args(argv=None):
                               " %(default)s seconds - increase this if you"
                               " are under a really slow network."))
 
+    parser.add_argument('--pref', nargs='*', dest='prefs',
+                        help=("A preference to set. Must be a key-value pair"
+                              "separated by a ':'"))
+
+    parser.add_argument('--preferences',
+                        help=("read preferences from a JSON or INI file. For"
+                              "INI, use 'file.ini:section' to specify a"
+                              "particular section."))
+
     commandline.add_logging_group(parser)
     options = parser.parse_args(argv)
     options.bits = parse_bits(options.bits)
@@ -239,6 +250,10 @@ def cli(argv=None):
     else:
         test_runner = CommandTestRunner(fetch_config, options.command,
                                         persist=options.persist)
+
+    prefs = mozprofile.prefs.Preferences()
+    if options.preferences:
+        prefs.add_file(options.preferences)
 
     runner = BisectRunner(fetch_config, test_runner, options)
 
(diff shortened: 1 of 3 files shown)


mozregression/main.py
```diff
@@ -240,6 +251,10 @@ def cli(argv=None):
         test_runner = CommandTestRunner(fetch_config, options.command,
                                         persist=options.persist)
 
+    prefs = mozprofile.prefs.Preferences()
+    if options.preferences:
+        prefs.add_file(options.preferences)
+
```
parkouss: you forgot to add preferences from the options.prefs variable

parkouss:
This looks good!
I just noted something on the PR, you forgot to handle the --pref argument.

parkouss:
Also, the pref object needs do be added here (in launcher_kwargs): https://github.com/mozilla/mozregression/pull/190/files#diff-9ca866adb100de14c577141d9b0d6b67R245, else it does not have any effect.

mozregression/main.py
imjalpreet: I didn't get this. Can you please explain this?

imjalpreet:
Do I have to add in the following way: 
prefs = options.pref ? or prefs = mozprofile.prefs.Preferences() ?

mozregression/main.py
parkouss: Well, prefs is our object that contain preferences. You added preferences from the --preferences file, but not from the --pref argument. Have a look at the implementation of the same thing for mozprofile: http://mozbase.readthedocs.org/en/latest/_modules/mozprofile/cli.html#MozProfileCLI.preferences

Also, could you create a function that will return the prefs object ? It will be easier to test. Last thing, could you make the --preferences a list (like for --pref) by adding nargs='*' ? I changed my mind on this after looking at the behaviour of mozprofile.

parkouss:
You have to add the created prefs object like this:

launcher_kwargs = dict(
addons=options.addons,
profile=options.profile,
cmdargs=options.cmdargs,
preferences=prefs,
)

This is because the variable launcher_kwargs will then be passed here : https://github.com/mozilla/mozregression/blob/master/mozregression/test_runner.py#L29, then we will use it here: https://github.com/mozilla/mozregression/blob/master/mozregression/test_runner.py#L131 and finally it will override the default that you just added here: https://github.com/mozilla/mozregression/pull/190/files#diff-9be3d66461b771098aa1bcc305505480R113

This is how your prefs object will end up in the launchers.MozRunnerLauncher._start method.

Tell me if it is not clear, I think it is pretty important to understand this.

imjalpreet:
@parkouss, I will get back to you by tonight (2-3 hours).

mozregression/main.py
imjalpreet: Sorry for the delay. I wanted to confirm that from the above link you want me to implement something similar to the preferences function, right? and secondly the --pref argument is to be implemented in a similar way as the CLI preferences have been implemented, right?

parkouss: Yep, that's right :) The idea is to be able to pass the arguments in the same way as in mozprofile, create a preferences object (and fill it accordingly to the passe arguments) then give that object to our launcher.

imjalpreet:
Due to some reasons, I am sending you a new pull request.

PR #192, opened 2015-02-24
#Bug 1106804 - --inbound-branch should imply --inbound

rahulrrixe:
As suggested by @wlach We want to bisect inbound if last good revision and first bad revision are set, so I added a condition for this. Let me know if this is the correct way to implement it.

### DIFF
--- a/mozregression/main.py
+++ b/mozregression/main.py
@@ -37,7 +37,7 @@ def parse_args(argv=None):
              " [[--good GOOD_DATE]|[--good-release GOOD_RELEASE]]"
              "\n"
              " %(prog)s [OPTIONS]"
-             " --inbound --bad-rev BAD_REV --good-rev GOOD_REV")
+             " --bad-rev BAD_REV --good-rev GOOD_REV")
 
     parser = ArgumentParser(usage=usage)
     parser.add_argument("--version", action="version", version=__version__,
@@ -69,18 +69,13 @@ def parse_args(argv=None):
                         help=("last known good nightly build. This option is"
                               " incompatible with --good."))
 
-    parser.add_argument("--inbound",
-                        action="store_true",
-                        help=("use inbound instead of nightlies (use"
-                              " --good-rev and --bad-rev options."))
-
     parser.add_argument("--bad-rev", dest="first_bad_revision",
-                        help=("first known bad revision (use with"
-                              " --inbound)."))
+                        help=("first known bad revision (for inbound"
+                              " bisection)."))
 
     parser.add_argument("--good-rev", dest="last_good_revision",
-                        help=("last known good revision (use with"
-                              " --inbound)."))
+                        help=("last known good revision (for inbound"
+                              " bisection)."))
 
     parser.add_argument("--find-fix", action="store_true",
                         help="Search for a bug fix instead of a regression.")
@@ -247,10 +242,12 @@ def cli(argv=None):
         # can go to inbound from nightly.
         fetch_config.set_inbound_branch(options.inbound_branch)
 
-    if options.inbound:
+    # bisect inbound if last good revision or first bad revision are set
+    if options.first_bad_revision or options.last_good_revision:
         bisect = bisect_inbound
     else:
         bisect = bisect_nightlies
+
     try:
         launcher_class = APP_REGISTRY.get(fetch_config.app_name)
         launcher_class.check_is_runnable()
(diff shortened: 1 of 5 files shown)


rahulrrixe:
@parkouss I have removed the --inbound flag.

mozregression/main.py
```diff
@@ -247,10 +240,12 @@ def cli(argv=None):
         # can go to inbound from nightly.
         fetch_config.set_inbound_branch(options.inbound_branch)
 
-    if options.inbound:
+    # bisect inbound if last good revision and first bad revision are set
+    if (options.first_bad_revision and options.last_good_revision):
```
parkouss: You have to use a "or" condition here instead of "and". If there is a missing rev option (--good-rev or --bad-rev) it wil be handled in the bisect_inbound function (https://github.com/mozilla/mozregression/blob/9a99d82071f383147b14899b64c96bad467e6bd6/mozregression/main.py#L217).
if you don't do that, a user that only use "mozregression --good-rev 123" for example will default to nightly bisection instead of inbound which is not really user-friendly.

tests/unit/test_main.py
```diff
@@ -68,20 +68,10 @@ def test_basic_inbound(self):
         # we exited with the return value of bisect_inbound
         self.assertEquals(exitcode, 0)
 
-    def test_inbound_revs_must_be_given(self):
```
parkouss: Instead of removing this test, you can just change the arglist variable by removing all the --inbound in it, eg:
argslist = [['--good-rev=1'], ['--bad-rev=5']]
The test must then pass again (given that you changed the "and" to an "or" like I said in my last comment)

rahulrrixe:
Hi @parkouss and @wlach, I have stashed the commits into one file. Can you please review it?

mozregression/main.py
```diff
@@ -69,18 +69,11 @@ def parse_args(argv=None):
...
 
-    parser.add_argument("--inbound",
-                        action="store_true",
-                        help=("use inbound instead of nightlies (use"
-                              " --good-rev and --bad-rev options."))
-
     parser.add_argument("--bad-rev", dest="first_bad_revision",
-                        help=("first known bad revision (use with"
-                              " --inbound)."))
+                        help=("first known bad revision"))
```
parkouss: Could you add something to indicate that we bisect inbound ?
Maybe "first know bad revision (for inbound bisection)"
Also, please remove the parentheses around the help text if it is a one line.

mozregression/main.py
```diff
@@ -69,18 +69,11 @@ def parse_args(argv=None):
...
 
     parser.add_argument("--good-rev", dest="last_good_revision",
-                        help=("last known good revision (use with"
-                              " --inbound)."))
+                        help=("last known good revision"))
```
parkouss: Same thing here.

mozregression/main.py
```diff
@@ -247,10 +240,12 @@ def cli(argv=None):
         # can go to inbound from nightly.
         fetch_config.set_inbound_branch(options.inbound_branch)
 
-    if options.inbound:
+    # bisect inbound if last good revision or first bad revision are set
+    if (options.first_bad_revision or options.last_good_revision):
```
parkouss: Could you remove the parentheses here please ? We don't need them.

parkouss:
I'm sorry @rahulrrixe, I missed some points in my previous review... Could you update the PR please ? After that I'll merge this in. Thanks. :)

rahulrrixe:
@parkouss I have updated the PR with recommended changes :+1: Also can you give me a link to documentation to read about these code formatting rules.

parkouss:
Thanks @rahulrrixe !
You can find most information in the pep8 recommendation guide: https://www.python.org/dev/peps/pep-0008/.
However, I don't see a word for the use or not of parenthesis around a single line if statement. Well simply most python developpers do not use them (as we do), so it is mainly to be consistent with the rest of our code. :)

parkouss:
Yay! I'm really sorry, I forgot one last thing! There is a remaining --inbound argument here:
https://github.com/parkouss/mozregression/blob/9a99d82071f383147b14899b64c96bad467e6bd6/mozregression/main.py#L67
Could you fix that please ?
I promise this is the last one :)

rahulrrixe:
@parkouss No need to say sorry. Its Ok I am learning new things :+1:

PR #194, opened 2015-03-02
Mozregression may ask user for upgrade

dangerouspython:
Added the code that checks for a new version of Mozregression when Mozrgression is started from the terminal/command prompt.
(Asking users for a upgrade might be beneficial in this case as better version of mozregression might be having improved features
 that the user can use for finding out the regression range in the corresponding mozilla products)

If a new version of Mozregression is out on PyPI, then Mozregression informs the
user about it and suggests the user to upgrade to the new version by issuing one
of the two commands:

		1) pip install -U mozregression
		2) pip install --upgrade mozregression

If the user already has the latest version, then Mozregression continues as usual.

### DIFF
--- a/mozregression/main.py
+++ b/mozregression/main.py
@@ -13,6 +13,7 @@ import datetime
 import sys
 import atexit
 import os
+import requests
 
 import mozprofile
 from argparse import ArgumentParser
@@ -286,6 +287,24 @@ def preference(prefs_files, prefs_args):
     return prefs()
 
 
+def check_mozregression_version(logger):
+    url = "https://pypi.python.org/pypi/mozregression/json"
+    try:
+        mozregression_version = \
+            requests.get(url, timeout=10).json()['info']['version']
+    except (RequestException, KeyError, ValueError):
+        logger.critical("Unable to get latest version from pypi.")
+        return
+
+    if __version__ != mozregression_version:
+        logger.warning("You are using mozregression version %s, "
+                       "however version %s is available."
+                       % (__version__, mozregression_version))
+
+        logger.warning("You should consider upgrading via the 'pip install"
+                       " --upgrade mozregression' command.")
+
+
 def cli(argv=None):
     """
     main entry point of mozregression command line.
@@ -294,6 +313,7 @@ def cli(argv=None):
     logger = commandline.setup_logging("mozregression",
                                        options,
                                        {"mach": sys.stdout})
+    check_mozregression_version(logger)
 
     if options.list_releases:
         print(formatted_valid_release_dates())


parkouss:
Hey @dangerouspython, you should rebase your changes to be in sync with the current master branch of mozregression. try a "git pull --rebase https://github.com/mozilla/mozregression", merge the conflicts if any, then you can push again your changes with a "git push -f" (that will rewrite the history of your fork).

mozregression/main.py
```diff
@@ -31,6 +33,7 @@
 from mozregression.launchers import REGISTRY as APP_REGISTRY
 from mozregression.test_runner import ManualTestRunner, CommandTestRunner
 
+LOG = getLogger(__name__.split('.')[0])
```
parkouss: You don't need this - you can just use the logger created in the cli() function.

mozregression/main.py
```diff
@@ -286,10 +289,39 @@ def preference(prefs_files, prefs_args):
     return prefs()
 
 
+def check_mozregression_version():
+    url = "https://pypi.python.org/pypi/mozregression/json"
+
+    LOG.info("Checking for availability of the"
```
parkouss: Hmm, I think it will just annoy users to see that each time they run the command - I think you can remove this line

mozregression/main.py
```diff
@@ -286,10 +289,39 @@ def preference(prefs_files, prefs_args):
...
+
+    if __version__ != mozregression_version:
+        LOG.warn("You are using mozregression version %s, however version %s"
+                 " is available.", __version__, mozregression_version)
+
+        LOG.warn("You should consider upgrading via the 'pip install"
+                 " --upgrade mozregression'\n or via"
+                 " 'pip install -U mozregression' command.")
+
+    else:
```
parkouss: Same here, you can just remove that.

mozregression/main.py
```diff
@@ -286,10 +289,39 @@ def preference(prefs_files, prefs_args):
...
+        LOG.info("\nYou already have the latest version! Continuing...\n")
+
+    return
```
parkouss: the return here is not needed.

dangerouspython:
@parkouss : Can you please review the bugfix?

mozregression/main.py
```diff
@@ -286,6 +287,24 @@ def preference(prefs_files, prefs_args):
...
+        logger.critical("Unable to get latest version from pypi.")
+        return
+
+    if __version__ != mozregression_version:
+        logger.warn("You are using mozregression version %s, "
+                    "however version %s is available.",
+                    __version__, mozregression_version)
+
+        logger.warn("You should consider upgrading via the 'pip install"
+                    " --upgrade mozregression'\n or via"
```
parkouss: I would prefer something more simple here. Could you remove the part "\n or via..." ? It is quite redundant.

PR #198, opened 2015-03-07
Unit tests for the function 'check_mozregression_version' in mozregression.main.py

dangerouspython:
It would be better to have some unit tests for the function 'check_mozregression_version' in mozregression.main.py. Thats why I wrote these functions

### DIFF
--- a/tests/unit/test_main.py
+++ b/tests/unit/test_main.py
@@ -9,9 +9,11 @@ from mock import patch, Mock
 import datetime
 import tempfile
 import os
+import requests
 
 from mozregression import main, utils, errors
 from mozregression.test_runner import CommandTestRunner
+from mozregression import __version__
 
 
 class TestMainCli(unittest.TestCase):
@@ -245,6 +247,32 @@ class TestPreference(unittest.TestCase):
 
         prefs = main.preference(None, prefs_args)
         self.assertEquals(len(prefs), 0)
+
+
+class TestCheckMozregresionVersion(unittest.TestCase):
+    @patch('requests.get')
+    def test_version_is_upto_date(self, get):
+        logger = Mock()
+        response = Mock(json=lambda: {'info': {'version':  __version__}})
+        get.return_value = response
+        main.check_mozregression_version(logger)
+        self.assertFalse(logger.critical.called)
+
+    @patch('requests.get')
+    def test_Exception_error(self, get):
+        logger = Mock()
+        get.side_effect = requests.RequestException
+        # exception is handled inside main.check_mozregression_version
+        main.check_mozregression_version(logger)
+        self.assertRaises(requests.RequestException, get)
+
+    @patch('requests.get')
+    def test_warn_if_version_is_not_up_to_date(self, get):
+        logger = Mock()
+        response = Mock(json=lambda: {'info': {'version': 0}})
+        get.return_value = response
+        main.check_mozregression_version(logger)
+        self.assertEqual(logger.warning.call_count, 2)
 
 if __name__ == '__main__':
     unittest.main()


tests/unit/test_main.py
```diff
@@ -246,5 +248,69 @@ def test_preference_args(self):
         prefs = main.preference(None, prefs_args)
         self.assertEquals(len(prefs), 0)
 
+
+class TestCheckMozregresionVersion(unittest.TestCase):
+    @patch('requests.get')
+    @patch('sys.stdout')
+    def test_pypi_is_accesible(self, stdout, get):
```
parkouss: Hmm, I don't understand the name of this function. Shouldn'it be something like "test_version_is_up_to_date" ?

tests/unit/test_main.py
```diff
@@ -246,5 +248,69 @@ def test_preference_args(self):
...
+        options = None
+        logger = Mock('mozlog.structured.commandline.setup_logging')(
```
parkouss: I think that just looger = Mock() is sufficient here. Plus, you won't have to mock stdout.

dangerouspython: Ok.. Sure! Will re-push it again in a few seconds...

tests/unit/test_main.py
```diff
@@ -246,5 +248,69 @@ def test_preference_args(self):
...
+    @patch('requests.get')
+    @patch('sys.stdout')
+    def test_RequestException_error(self, stdout, get):
+        options = None
+        logger = Mock('mozlog.structured.commandline.setup_logging')(
+                      "mozregression",
+                      options,
+                      {"mach": stdout})
+        get.side_effect = requests.RequestException
+        main.check_mozregression_version(logger)
```
parkouss: could you add a comment above this to indicate that the exception is handled inside main.check_mozregression_version ?

tests/unit/test_main.py
```diff
@@ -246,5 +248,69 @@ def test_preference_args(self):
...
+        self.assertRaises(requests.RequestException, get)
+
+    @patch('requests.get')
+    @patch('sys.stdout')
+    def test_key_error(self, stdout, get):
```
parkouss: Well, this seems not really useful to me. the code executed is the same for the previous test. If you want to test every kind of handled exception, maybe write a loop in one test method. or jut get rid of this test method :)

tests/unit/test_main.py
```diff
@@ -246,5 +248,69 @@ def test_preference_args(self):
...
+                      "mozregression",
+                      options,
+                      {"mach": stdout})
+        get.side_effect = KeyError
+        main.check_mozregression_version(logger)
+        self.assertRaises(KeyError, get)
+
+    @patch('requests.get')
+    @patch('sys.stdout')
+    def test_value_error(self, stdout, get):
```
parkouss: Same as above I think.

tests/unit/test_main.py
```diff
@@ -246,5 +248,69 @@ def test_preference_args(self):
...
+                      "mozregression",
+                      options,
+                      {"mach": stdout})
+        get.side_effect = ValueError
+        main.check_mozregression_version(logger)
+        self.assertRaises(ValueError, get)
+
+    @patch('requests.get')
+    @patch('sys.stdout')
+    def test_warning_is_called_twice(self, stdout, get):
```
parkouss: I don't really like the name here. Maybe "test_warn_if_version_not_up_to_date" ?

parkouss:
Hey @dangerouspython, thanks for this ! I noted some thing on the PR, could you address them please ?

PR #201, opened 2015-03-16
Gui panel header to replace wizard

parkouss:


parkouss:
Closing this for now - we'll see later if you want it.

PR #211, opened 2015-04-23
fix default to a one year regression range instead of year 2000 

MikeLing:


parkouss:
This is a PR created by mistake, so I close it.

PR #212, opened 2015-04-25
error checking for the regression range

MikeLing:
Add the error check before user submit.

### DIFF
--- a/gui/mozregui/wizard.py
+++ b/gui/mozregui/wizard.py
@@ -111,6 +111,26 @@ class NightliesPage(WizardPage):
         self.ui.start_date.setDateTime(now.addYears(-1))
         self.ui.end_date.setDateTime(now)
 
+    def validatePage(self):
+        start_date = self.ui.start_date.date()
+        end_date = self.ui.end_date.date()
+        current = QDateTime.currentDateTime().date()
+        if start_date < end_date:
+            if end_date <= current:
+                return True
+            else:
+                QMessageBox.critical(
+                    self,
+                    "Error",
+                    "You can't define a date in the future.")
+        else:
+            QMessageBox.critical(
+                self,
+                "Error",
+                "The first date must be earlier than the second one.")
+
+        return False
+
     def nextId(self):
         return ProfilePage.ID
 


gui/mozregui/wizard.py
```diff
@@ -111,6 +111,27 @@ def __init__(self):
         self.ui.start_date.setDateTime(now.addYears(-1))
         self.ui.end_date.setDateTime(now)
 
+    def validatePage(self):
+        good_date = self.ui.start_date.date()
+        bad_date = self.ui.end_date.date()
```
parkouss: Could you rename the variables here as "start_date" and "end_date" ? Because this code also handle the find fix feature (so good and bad will means the exact the opposite, which will be confusing) - but the start / end terminology will be good in any case. :)

MikeLing: Sure! absolutely:)

gui/mozregui/wizard.py
```diff
@@ -111,6 +111,27 @@ def __init__(self):
...
+        current = QDateTime.currentDateTime().date()
+        if good_date < bad_date:
+            if bad_date <= current:
+                return True
+            else:
+                QMessageBox.critical(
+                    self,
+                    "Error",
+                    "Date you entered is larger than now,please try it again.")
```
parkouss: I'm not sure the message here is great english (I'm not a natural english speaker neither but larger for a date seems strange to me). Also I find the "please try again" not really useful.
Maybe "You can't define a date in the future." would be enough ?

gui/mozregui/wizard.py
```diff
@@ -111,6 +111,27 @@ def __init__(self):
...
+                return False
+        else:
+            QMessageBox.critical(
+                self,
+                "Error",
+                "Dates in good date field can't larger than bad date,\
+                please try it again.")
```
parkouss: Same thing for the "please try", not sure it helps. Also we should avoid to use the good / bad words here (since user may have checked "Search fix"). Maybe "The first date must be earlier than the second one." ?

Last thing to note, When a line with a string is too long, please use something like here: https://github.com/mozilla/mozregression/blob/master/mozregression/main.py#L93. The syntax you used is not equivalent, as it add spaces into the string.

gui/mozregui/wizard.py
```diff
@@ -111,6 +111,27 @@ def __init__(self):
...
+            return False
```
parkouss: Could you put the "return False" at the end of the function, and remove the two others ? I think it will be more readable.

gui/mozregui/wizard.py
```diff
@@ -146,7 +167,6 @@ def __init__(self, parent=None):
         QWizard.__init__(self, parent)
         self.setWindowTitle("Bisection wizard")
         self.resize(800, 600)
-
```
parkouss: Hmm, this is not required in this patch, please remove it.

gui/mozregui/wizard.py
MikeLing: Sure :)

Actually, I wonder if the message there will make people confuse when I write it. Hmm, maybe I can improve many other things except coding if I continue to work on open source. :D

PR #213, opened 2015-04-26
fix make the "search fix" feature works as expected 

MikeLing:
make the "search fix" feature works as expected

### DIFF
--- a/gui/mozregui/bisection.py
+++ b/gui/mozregui/bisection.py
@@ -237,12 +237,12 @@ class BisectRunner(QObject):
         self.bisector_created.emit(self.bisector)
         if options['bisect_type'] == 'nightlies':
             handler = NightlyHandler(find_fix=options['find_fix'])
-            start = options['start_date']
-            end = options['end_date']
+            good = options['good_date']
+            bad = options['bad_date']
         else:
             handler = InboundHandler(find_fix=options['find_fix'])
-            start = options['start_changeset']
-            end = options['end_changeset']
+            good = options['good_changeset']
+            bad = options['bad_changeset']
 
         # options for the app launcher
         launcher_kwargs = {}
--- a/gui/mozregui/wizard.py
+++ b/gui/mozregui/wizard.py
@@ -99,22 +99,42 @@ class IntroPage(WizardPage):
             return InboundPage.ID
 
 
-class NightliesPage(WizardPage):
+class WizardSelectionRangePage(WizardPage):
+    def changelabel(self, checkstatus):
+        if checkstatus is True:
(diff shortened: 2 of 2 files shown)


gui/mozregui/bisection.py
```diff
@@ -237,12 +237,12 @@ def bisect(self, fetch_config, options):
         self.bisector_created.emit(self.bisector)
         if options['bisect_type'] == 'nightlies':
             handler = NightlyHandler(find_fix=options['find_fix'])
-            start = options['start_date']
-            end = options['end_date']
+            start = options['good_date']
+            end = options['bad_date']
```
parkouss: Could you also remove the variable start and end to good and bad here ?

gui/mozregui/bisection.py
```diff
@@ -237,12 +237,12 @@ def bisect(self, fetch_config, options):
...
         else:
             handler = InboundHandler(find_fix=options['find_fix'])
-            start = options['start_changeset']
-            end = options['end_changeset']
+            start = options['good_changeset']
+            end = options['bad_changeset']
```
parkouss: Same as above.

gui/mozregui/wizard.py
```diff
@@ -105,6 +114,40 @@ class NightliesPage(WizardPage):
     FIELDS = {"start_date": "start_date", "end_date": "end_date"}
     ID = 1
 
+<<<<<<< HEAD
```
parkouss: Hmm, something's wrong here. :)

gui/mozregui/wizard.py
```diff
@@ -29,6 +29,15 @@ def resolve_obj_name(obj, name):
     return obj
 
 
+def changelabel(self, checkstatus):
```
parkouss: Could you make this a method in a class ? I think it would be easier to understand and better structured.
Maybe in a:

class WizardSelectionRangePage(WizardPage):
    def change_label()
        ...
    def initializePage()
        ...
Then you could inherit from it for inbound/Nightlies:

class InboundPage(WizardSelectionRangePage):
    ...
class NightliesPage(WizardSelectionRangePage):
    ...

MikeLing: Great! That's a good idea!

gui/mozregui/wizard.py
MikeLing: So, should I commit, rebase, push -f after I fix the probelm this time?

parkouss: Yep, I think this will do the job - commit rebase and push -f. Be sure you only make a patch that do what you intend here (and not some other stuff that comes from the other branch).

PR #215, opened 2015-04-29
fix bug 1158125

MikeLing:
Hmm, I hope it may works this time:)

### DIFF
--- a/gui/mozregui/wizard.py
+++ b/gui/mozregui/wizard.py
@@ -109,7 +109,21 @@ class IntroPage(WizardPage):
             return InboundPage.ID
 
 
-class NightliesPage(WizardPage):
+class WizardSelectionRangePage(WizardPage):
+    def changelabel(self, checkstatus):
+        if checkstatus is True:
+            self.ui.label.setText("Last known bad date")
+            self.ui.label_2.setText("First known good date")
+        else:
+            self.ui.label.setText("Last known good date")
+            self.ui.label_2.setText("First known bad date")
+
+    def initializePage(self):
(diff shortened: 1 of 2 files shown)


gui/mozregui/wizard.py
```diff
@@ -145,13 +159,16 @@ def nextId(self):
...
 
-class InboundPage(WizardPage):
+class InboundPage(WizardSelectionRangePage):
     UI_CLASS = Ui_Inbound
     TITLE = "Select the inbound changesets range"
     FIELDS = {"start_changeset": "start_changeset",
               "end_changeset": "end_changeset"}
     ID = 2
 
+    def initializePage(self):
```
parkouss: Could you remove this ? This is not required, the base method will be called anyway.

gui/mozregui/wizard.py
```diff
@@ -198,6 +215,20 @@ def options(self):
                 elif isinstance(value, QDate):
                     value = value.toPyDate()
                 options[fieldname] = value
+        if options['bisect_type'] == 'nightlies':
```
parkouss: Hmm looking at this code, it seems a bit redundant. Maybe you could refactorise this a bit, for example:
kind = "date" if options['bisect_type'] == 'nightlies' else "changeset"
if not options['find_fix']:
    options['good_' + kind ] = options.pop('start_' + kind)
    options['bad_' + kind] = options.pop('end_' + kind)
else:
    options['good_' + kind] = options.pop('end_' + kind)
    options['bad_' + kind] = options.pop('start_' + kind)

This way we don't duplicate the logic.

PR #219, opened 2015-05-14
fix bug 1161542 create a script that can freeze and create the instal…

MikeLing:


gui/build.py
```diff
@@ -79,6 +80,20 @@ def do_bundle():
     for dirname in ('build', 'dist'):
         if os.path.isdir(dirname):
             shutil.rmtree(dirname)
+    # create a intaller
+    env = dict(os.environ)
+    p = re.search('D:\\\\Python27\\\\Lib', env['PYTHONPATH'])
```
parkouss: Hm, not sure to understand what you are trying to do here...
I think that the python path "D:\\\\Python27" must be the configurable part. For this, you can do something like
python_lib_path = os.path.join(
    os.getenv('MOZREGUI_PYTHONPATH', "C:\\\\Python27"),
    "Lib"
)
Thiw way one can call this on the windows shell:
set MOZREGUI_PYTHONPATH=D:\Python27
python build.by bundle

If you prefer, instead of an environment variable you can use an option to the command line.

gui/build.py
```diff
@@ -79,6 +80,20 @@ def do_bundle():
...
+    python_path = p.group(0)
+    gui_path = os.path.realpath('.')
+    moz_path = os.path.realpath('..')
+    include_path = '--include-path=%s;%s;%s//site-packages;%s ' % (gui_path, moz_path, python_path, python_path)
+    icon_arg = '--icon=wininst/app_icon.ico '
+    target_arg = '--target-name=mozregression-gui.exe --target-dir=dist'
+    base_name = '--base-name=Win32GUI '
+    arg = icon_arg + base_name + include_path + target_arg
+    call(sys.executable, py_script('cxfreeze'), arg, 'mozregui/main.py')
```
parkouss: call() require you to list the arguments one by one - that is:
call('cd', 'mozregui')
but you can put them in a list and then use * (http://stackoverflow.com/questions/3394835/args-and-kwargs)
args = ['cd', 'mozregui']
call(*args)

So here you should build args this way for example:
args = []
args.append('--icon=wininst/app_icon.ico')  # note: no leading space
args.append('--base-name=Win32GUI')
...
args.append('mozregui/main.py')
call(sys.executable, py_script('cxfreeze'), *args)

gui/build.py
```diff
@@ -79,6 +80,20 @@ def do_bundle():
...
+    os.system('cd wininst')
+    os.system('makensis wininst.nsi')
```
parkouss: you should make use of call() here too (that calls subprocess) in this way:
call('makensis', 'wininst.nsi', cwd='wininst')

cwd named argument is explained here: https://docs.python.org/2/library/subprocess.html#popen-constructor.

gui/build.py
MikeLing: Hm, I remember you tell me on irc that I should get python path on users' system rather than path in virtual environment. So, I get my os environment path and pick python path from them. But env['PYTHONPATH'] give me a string contain three python path('D:\\Python27\\Lib;D:\\Python27\\bin;D:\\Python27\\Lib\\Script').
I though I could use regular expression to pick only one path from it, but the problem is not everyone install python on disk 'D' or maybe the python path is 'D:\xxxx\xxxx\Python27' and I don't know how to solve it.

parkouss: This is why we have to use an environment var (or a command line flag). We need the Python system path for the cxfreeze command, but we can't discover it automatically it seems. So the idea is to define it with a default variable (I would say "C:\Python27" - as this is the most commn default I think) and let the user change it at runtime - by setting the environment var or using a command line flag.

gui/build.py
```diff
@@ -79,6 +79,28 @@ def do_bundle():
     for dirname in ('build', 'dist'):
         if os.path.isdir(dirname):
             shutil.rmtree(dirname)
+    # create a intaller
+    python_lib_path = os.path.join(
+        os.getenv('MOZREGUI_PYTHONPATH', "C:\\Python27"),
+        "Lib"
+    )
+    makensis_path = os.path.join(
+        os.getenv('MOZREGUI_PYTHONPATH', "C:\\NSIS"),
```
parkouss: The env var name is wrong here - maybe 'MOZREGUI_NSISPATH' ?

gui/build.py
```diff
@@ -79,6 +79,28 @@ def do_bundle():
...
+    args.append('--include-path=%s;%s;%s\site-packages;%s'
+                % (gui_path, moz_path, python_lib_path, python_lib_path))
+    args.append('--target-name=mozregression-gui.exe')
+    args.append('--target-dir=dist')
+    args.append('mozregui/main.py')
+    call(sys.executable, os.path.join(python_dir, 'cxfreeze'), *args)
+    call(makensis_path, 'wininst.nsi', cwd='wininst')
 
     # run pyinstaller
     pyinstaller = py_script('pyinstaller')
```
parkouss: could you remove pyinstaller and the call to pyintsaller ? cxfreeze is replacing pyinstaller here.

parkouss:
Also, could you try to adapt the appveyor.yml file to use this new "python build.py bundle" command ?

Thanks @MikeLing !

gui/build.py
MikeLing: Hm, It's my fault for forget to remove it. Actually, I had change the code when I work under the windows but I need switch to Linux and change my code again before I commit and push them.
So, a lot mistakes happened because of my carelessness.....:(

MikeLing:
I'm wondering if this line has use "python build.py bundle" in appveyor.yml?

https://github.com/MikeLing/mozregression/blob/master/appveyor.yml#L24

parkouss:
In fact it seems that your for is a bit out of date (this line don't exist anymore in the code):
https://github.com/mozilla/mozregression/blob/master/appveyor.yml#L22
You should update it, for example with "git pull --rebase".

parkouss:
Thanks! I manually merged this: https://github.com/mozilla/mozregression/commit/8c7e04b8b8a90141d73af79a0807414bf0c3a3b7

PR #221, opened 2015-05-16
Misc gui

jpigree:
Multiple little tweaks to mozregression-gui.
In details:
- Add ArchLinux PyQt dev tools install steps
- Add subtitles to each wizard's pages
- Fix mozregression-gui name in ctrl-tab menu in gnome3

### DIFF
--- a/gui/README.rst
+++ b/gui/README.rst
@@ -20,9 +20,13 @@ tasks (as some replacement for the standard setup.py).
 First thing is to install pyqt4. This must be done system wide (PyQT4
 is not installable via pip).
 
-- On ubuntu: ::
+- On Ubuntu: ::
 
     sudo apt-get install python-qt4 pyqt4-dev-tools
+
+- On Archlinux: ::
+
+  sudo pacman -S python2-pyqt4
 
 - On OSX (using MacPorts_): ::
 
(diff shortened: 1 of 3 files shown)


jpigree:
Okay. I forced push to add your new changes on mozilla/master + the minor tweaks of the gui doc.

jpigree:
Yeah, you are right. It seems to be a Qt Bug. I renamed it also because it seemed to me that an executable named with the application name was better than a main.py which is more a developer thing (main + .py).

I guess it is just my point of view. I understand that change the entry point like that is way too more impactful for such a minor bug.

I will try out your solution which I guess will work as a workaround.

jpigree:
Needs some tests on windows though.

parkouss:
Oh, I'm sorry I thought that you wanted to rename the whole application name. :) Now I get that it was only the file name that you wanted to rename!
Well I'm not against renaming that file, but it seemed that the cx freeze did not like it - maybe we can investigate this. Or we can just tweak the argv given to the application (maybe just removig the '.py' final part ?, or something like suggested above) and it will be good - I don't think that Qt cares about the file name (but maybe it cares about the file path).

parkouss:
Awesome, thanks @jpigree !

PR #223, opened 2015-05-17
Gui tests

parkouss:
This add a simple test for the persist folder that we just added.


parkouss:
@jpigree What do you think of it ? I'd like a review if you have the time since you implemented the persist gui option. :)

PR #224, opened 2015-05-18
Add some options for bundle when create the installer on windows

MikeLing:


gui/build.py
```diff
@@ -80,14 +80,21 @@ def do_bundle(options):
         if os.path.isdir(dirname):
             shutil.rmtree(dirname)
     # create a intaller
+    if IS_WIN:
```
parkouss: We don't need the IS_WIN check yet - for now this part of the script is windows only (we will make it work for linux with the next bug). So please remove it for now. Also, the default values for python_path and nsis_path can be defined as a default= argument in the add_argument call, so you can just remove all this block.

gui/build.py
```diff
@@ -124,6 +131,10 @@ def parse_args():
 
     bundle = subparsers.add_parser('bundle',
                                    help='bundle the application (freeze)')
+    bundle.add_argument('--python-path', dest='python_path',
+                        help='your python path on system')
+    bundle.add_argument('--nsis-path', dest='nsis_path',
+                        help='your nsis path on system')
```
parkouss: please put the default values here, it will be simpler.

gui/build.py
```diff
@@ -80,14 +80,21 @@ def do_bundle(options):
...
+        if options.python_path:
+            python_path = options.python_path
+        if options.nsis_path:
+            nsis_path = options.nsis_pat
     python_lib_path = os.path.join(
-        os.getenv('MOZREGUI_PYTHONPATH', "C:\\Python27"),
+        os.getenv('MOZREGUI_PYTHONPATH', python_path),
         "Lib"
-    )
+        )
```
parkouss: you can let the parenthesis where they were, it was fine.

gui/build.py
```diff
@@ -80,14 +80,21 @@ def do_bundle(options):
...
+    if IS_WIN:
+        python_path = 'C:\\Python27'
+        nsis_path = 'C:\\NSIS'
+        if options.python_path:
+            python_path = options.python_path
+        if options.nsis_path:
+            nsis_path = options.nsis_pat
     python_lib_path = os.path.join(
-        os.getenv('MOZREGUI_PYTHONPATH', "C:\\Python27"),
+        os.getenv('MOZREGUI_PYTHONPATH', python_path),
```
parkouss: No need for the os.getenv now - we are replacing this with options.nsis_path.

gui/build.py
```diff
@@ -80,14 +80,21 @@ def do_bundle(options):
...
         "Lib"
-    )
+        )
     makensis_path = os.path.join(
-        os.getenv('MOZREGUI_NSISPATH', "C:\\NSIS"),
+        os.getenv('MOZREGUI_NSISPATH', nsis_path),
         "makensis.exe"
-    )
+        )
```
parkouss: same thing here - let the parenthesis where they were, and please remove the os.getenv call.

gui/build.py
MikeLing: Hm, I see. But when we call bundle under linux, the default argument will be also given to bundle if we forget to give a python path. So, why don't we set different default value of path when we call different bundle funtion?

parkouss: Well, we will! But it is for the next patch - for now, the bundle command is only used in windows, so let keep things simple.

MikeLing: I mean you can treat this patch as a windows bundle function.

> python_path = 'C:\\Python27'
> nsis_path = 'C:\\NSIS' 
And, we when bundle be called under Linux system.

> python_path = /usr/lib/python2.7

parkouss: The goal of this patch is to remove the environment variables used to use command line arguments instead. Let's just do this for now, and not something else - it is easy to do, and easy to review. Let's talk on irc about that!

MikeLing: I see. :)

Sorry, I have some classes this afternoon. I will send you email after that. Let's fix this issue first. :D

PR #225, opened 2015-05-19
Bug 1165566 - [gui] update build.py bundle command to make it works for linux

MikeLing:


gui/build.py
```diff
@@ -79,25 +81,37 @@ def do_bundle(options):
...
-    args.append('--base-name=Win32GUI')
-    args.append('--include-path=%s;%s;%s\site-packages;%s'
-                % (gui_path, moz_path, python_lib_path, python_lib_path))
-    args.append('--target-name=mozregression-gui.exe')
-    args.append('--target-dir=dist')
-    args.append('mozregui/main.py')
-    call(sys.executable, os.path.join(python_dir, 'cxfreeze'), *args)
-    call(makensis_path, 'wininst.nsi', cwd='wininst')
+    # create a installer
+    if sys.platform == 'win32':
```
parkouss: you should use the IS_WIN variable globally instead of testing sys.platform.

parkouss: Could you write a function call_cx_freeze ? I think that it will make the code a lot more readable. Something like:

def call_cx_freeze(gui_path, moz_path, python_lib_path):
    ...

In this function, you should taken into account the win/linux specificities, so the caller of the function
can just call the function without testing the OS.

gui/build.py
```diff
@@ -79,25 +81,37 @@ def do_bundle(options):
...
+        args.append('--target-name=mozregression-gui.exe')
+        args.append('--target-dir=dist')
+        args.append('mozregui/main.py')
+        call(sys.executable, os.path.join(python_dir, 'cxfreeze'), *args)
+        call(makensis_path, 'wininst.nsi', cwd='wininst')
+    elif sys.platform == 'linux2':
+        python_lib_path = options.python_path
+        gui_path = os.path.realpath('.')
+        moz_path = os.path.realpath('..')
+        os.system('cxfreeze --include-path="%s:%s:%s:%s/dist-packages"'
```
parkouss: You should use our call() function instead of os.system, because it is more robust for arguments that could contains spaces, and the exit code of the subprocess is tested.

gui/build.py
```diff
@@ -118,14 +132,20 @@ def parse_args():
...
+        bundle.add_argument('--python-path', dest='python_path',
+                            default='C:\\Python27',
+                            help='your python path on the'
+                            ' system(default: %(default)r)')
+        bundle.add_argument('--nsis-path', dest='nsis_path',
+                            default='C:\\NSIS',
+                            help='your NSIS path on the'
+                            ' system(default: %(default)r)')
+    elif sys.platform == 'linux2':
+        bundle.add_argument('--python-path', dest='python_path',
```
parkouss: instead of copy/pasting the whole option definition, you should just define a default variable that is different on each os. Something like:
if IS_WIN:
   default_python_path = 'C:\\Python27'
else:
   default_python_path = '/usr/lib/python2.7'

bundle.add_argument('--python-path', default=default_python_path, ...)

parkouss:
Thanks @MikeLing ! Globally this is fine - I would like some things to be refactorized, to avoid copy paste and make things easier to understand.

MikeLing:
Thank you for your advices :)
I hope this version can be more appropriate.

gui/build.py
```diff
@@ -79,25 +81,20 @@ def do_bundle(options):
...
-    args.append('--base-name=Win32GUI')
-    args.append('--include-path=%s;%s;%s\site-packages;%s'
-                % (gui_path, moz_path, python_lib_path, python_lib_path))
-    args.append('--target-name=mozregression-gui.exe')
-    args.append('--target-dir=dist')
-    args.append('mozregui/main.py')
-    call(sys.executable, os.path.join(python_dir, 'cxfreeze'), *args)
-    call(makensis_path, 'wininst.nsi', cwd='wininst')
+    if IS_WIN:
+        python_lib_path = os.path.join(options.python_path, "Lib")
```
parkouss: Hmm, I think we should set this in the default '--python-path':
default='C:\\Python27\\Lib'
This way we won't have to add Lib for windows. Thinking more about it, I think that we should rename the option to '--python-lib-path', because it is the lib path of python (note that for linux it is already the case: '/usr/lib/python2.7')
Also the variables in the code that are related should be renamed from python_path to python_lib_path.

Sorry for seeing this now, but I think it makes more sense. :)

gui/build.py
```diff
@@ -79,25 +81,20 @@ def do_bundle(options):
...
+        makensis_path = os.path.join(options.nsis_path, "makensis.exe")
+        call_cx_freeze(gui_path, moz_path, python_lib_path)
```
parkouss: Once the recommendation above will be done, this call_cx_freeze could be executed for both win and Linux - without testing for the os (outside the "if").

gui/build.py
```diff
@@ -79,25 +81,20 @@ def do_bundle(options):
...
+        call(makensis_path, 'wininst.nsi', cwd='wininst')
+    else:
+        python_lib_path = options.python_path
+        call_cx_freeze(gui_path, moz_path, python_lib_path)
+        release_tar = tarfile.open('mozregression-gui.tar.gz', 'w:gz')
```
parkouss: A good pythonic alternative is to use the "with statement" (http://effbot.org/zone/python-with-statement.htm). It is safer because the tar file will be closed even if there is an issue - and it is shorter to write. Something like:
with tarfile.open('mozregression-gui.tar.gz', 'w:gz') as release_tar:
    release_tar.add(r'dist/')

You can look at the example section and search for the one using with: https://docs.python.org/2/library/tarfile.html#examples

gui/build.py
```diff
@@ -118,19 +115,48 @@ def parse_args():
...
+                    % (gui_path, moz_path, python_lib_path, python_lib_path))
+        args.append('--target-name=mozregression-gui.exe')
+    else:
+        args.append('--include-path="%s:%s:%s:%s/dist-packages"'
+                    % (gui_path, moz_path, python_lib_path, python_lib_path))
+        args.append('--target-name=mozregression-gui')
+    args.append('--target-dir=dist')
+    args.append('mozregui/main.py')
+    if IS_WIN:
+        python_dir = os.path.join(
```
parkouss: I think it would be best to move this in the previous IS_WIN statement. The idea is that you can add all what you needs in args, then at the end just issue a:
call(*args)

gui/build.py
```diff
@@ -118,19 +115,48 @@ def parse_args():
...
+            os.path.dirname(sys.executable),
+            "Scripts"
+        )
+        call(sys.executable, os.path.join(python_dir, 'cxfreeze'), *args)
+    else:
+        call(py_script('cxfreeze'), *args)
```
parkouss: Same thing as above. I would prefer the call(*args) to be made outside the if statement. basically:
args = []
if IS_WIN:
   # add some args for win
else:
  # add some args for linux
# add common args
call(*args)

gui/build.py
MikeLing: Should I add all the arguments in args? For example, the '--include-path' under windows and Linux are different.  I think add two different --include-path at once may cause some problem.

gui/build.py
MikeLing: Is that means I should keep the IS_WIN? :)

parkouss:
Yes, this is a lot better! There is some great stuff in here. :)
Some last things to update and we should be good.

Oh, also (I almost forget!), could you update the .travis.yml file to make it use the python build.py bundle command ? This section needs to be updated: https://github.com/mozilla/mozregression/blob/master/.travis.yml#L34

Thanks !

gui/build.py
parkouss: Yes, but only one IS_WIN check is required - let's talk on irc if you have the time. :)

parkouss:
This looks good - A few minor last things that I'm going to write as github comments. Thanks @MikeLing for all your efforts here!

.travis.yml
```diff
@@ -36,8 +36,7 @@ before_deploy:
     - cd gui
```
parkouss: You can just issue the following command:
```
python gui/build.py bundle
```
Because the uic and rcc commands are done anyway when you use the bundle command (https://github.com/mozilla/mozregression/blob/master/gui/build.py#L75). That will allow you to remove the "cd " commands also.

MikeLing:
You are welcome!  :)

.travis.yml
MikeLing: I see :)

gui/build.py
```diff
@@ -79,25 +81,18 @@ def do_bundle(options):
...
-                % (gui_path, moz_path, python_lib_path, python_lib_path))
-    args.append('--target-name=mozregression-gui.exe')
-    args.append('--target-dir=dist')
-    args.append('mozregui/main.py')
-    call(sys.executable, os.path.join(python_dir, 'cxfreeze'), *args)
-    call(makensis_path, 'wininst.nsi', cwd='wininst')
+    python_lib_path = options.python_path
+    if IS_WIN:
+        makensis_path = os.path.join(options.nsis_path, "makensis.exe")
+        call_cx_freeze(gui_path, moz_path, python_lib_path)
```
parkouss: Since now call_cx_freeze needs to be done on the two platforms, could you move that line before the if IS_WIN ?
Also you won't need the python_lib_path variable (it will be used only once), just:
```
call_cx_freeze(gui_path, moz_path, options.python_lib_path)
```

gui/build.py
```diff
@@ -118,19 +113,48 @@ def parse_args():
...
+                            default='C:\\NSIS',
+                            help='your NSIS path on the'
+                            ' system(default: %(default)r)')
+    else:
+        default_python_path = '/usr/lib/python2.7'
+
     bundle.add_argument('--python-path', dest='python_path',
-                        default='C:\\Python27',
+                        default=default_python_path,
                         help='your python path on the'
```
parkouss: Please change this as help='your python library path on the'.

gui/build.py
```diff
@@ -118,19 +113,48 @@ def parse_args():
...
-    bundle.add_argument('--nsis-path', dest='nsis_path',
-                        default='C:\\NSIS',
-                        help='your NSIS path on the'
-                        ' system(default: %(default)r)')
     bundle.set_defaults(func=do_bundle)
 
     return parser.parse_args()
 
 
+def call_cx_freeze(gui_path, moz_path, python_lib_path):
```
parkouss: I' thinking that gui_path and moz_path may not need to be argument. They are always '.' and '..' (or the equivalent realpath). Maybe you could remove these arguments, and hardcode them like:
```
args.append('--include-path=.;..;%s\site-packages;%s'
```
This would cleanup the code a bit.

gui/build.py
```diff
@@ -118,19 +113,48 @@ def parse_args():
...
+    if IS_WIN:
+        default_python_path = 'C:\\Python27\\Lib'
+        bundle.add_argument('--nsis-path', dest='nsis_path',
+                            default='C:\\NSIS',
+                            help='your NSIS path on the'
+                            ' system(default: %(default)r)')
+    else:
+        default_python_path = '/usr/lib/python2.7'
+
     bundle.add_argument('--python-path', dest='python_path',
```
parkouss: please use '--python-lib-path' here. also the dest option is not needed (same thing for nsis_path option): https://docs.python.org/3/library/argparse.html#dest

gui/build.py
MikeLing: Ok, I will try it.
Ah, one more thing, would you mind offer the *args document to me again? I found it's very interesting and powerful and I would like to read it in detail  :)
Thank you very much!

parkouss:
Ok, so it appears that it does not freeze well the application (for linux and windows).
I fixed the patch then merged manually as f346e748025f7d100a2e2ba661028ddbac3bd320.
I'll write more information in https://bugzilla.mozilla.org/show_bug.cgi?id=1165566.

PR #226, opened 2015-05-21
use the filecache extra dependency with cachecontrol

parkouss:
I've made a patch for cachecontrol to easily track dependencies when using the local file cache (as we do):

https://github.com/ionrock/cachecontrol/commit/57dddb23ff220e4bac2df5ad5ff5df8304098ff5

This has now been released, so we should use it (this is called extra dependencies for pip).

@wlach @jpigree  r?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,9 +35,7 @@ setup(name="mozregression",
           'mozversion >= 1.2',
           'requests >= 2.5.0',
           'redo',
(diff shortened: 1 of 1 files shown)


jpigree:
Sorry I just saw your PR. Seems right to me. Also, I just updated my master and it works.

PR #228, opened 2015-05-31
Use a nice progressbar to report downloading of build files

jpigree:


jpigree:
Some line too long errors. :/ I am working on it.

jpigree:
The coverage tool will surely throw a warning since I didn't write unitary tests.

gui/mozregui/report_delegate.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class ReportItemDelegate(QStyledItemDelegate):
+    def __init__(self, parent=None, *args):
+        QStyledItemDelegate.__init__(self, parent, *args)
+        self.progress = 0
+        self.lastIndex = None
+
+    @Slot(object, int, int)
+    def show_dl_progress(self, dl, current, total):
```
parkouss: I feel like this logic should not be in the delegate, but in the model instead. That's the model job to keep data, and to signal that data has changed (model here is either the ReportModel object, or one of the ReportItem instance).

gui/mozregui/report_delegate.py
```diff
@@ -0,0 +1,51 @@
...
+        self.progress = (current * 100) / total
+        if self.lastIndex:
+            self.lastIndex.model().dataChanged.emit(
+                self.lastIndex, self.lastIndex)
+
+    def paint(self, painter, option, index):
+        # Write status
+        QStyledItemDelegate.paint(self, painter, option, index)
+
+        if not index.row() == (index.model().rowCount() - 1):
```
parkouss: I feel like we should ask the reportitem instance here if it is downloading. Maybe the reportitem could also store the percent information. Or it is possible to store this information for now on the ReportModel directly, as we are downloading only one at a time.

gui/mozregui/report_delegate.py
jpigree: you are completly right. I will fix it soon.

gui/mozregui/report_delegate.py
```diff
@@ -0,0 +1,51 @@
...
+            return
+
+        # Check if download is on
+        if "Downloading" in str(index.data().toPyObject()):
```
parkouss: Same as above. Note that maybe others signals could help to detect start/end of downloading: https://github.com/mozilla/mozregression/blob/master/gui/mozregui/bisection.py#L20

jpigree:
Okay thanks for the tips.You are right. I shouldn't base my detection of a download on some label which is prone to be changed. And the progress should be provided by the model.

gui/mozregui/report.py
```diff
@@ -60,11 +61,33 @@ def status_text(self):
         return msg % self.state_text
 
 
+def bulk_action_slots(slots, signal_object, slot_object, action):
```
parkouss: hmm, I would put the action argument at first. This way you could remove bulk_connect_slots and bulk_disconnect_slots:
```
bulk_action_slots('connect', slots, signal_object, slot_object)
bulk_action_slots('disconnect', slots, signal_object, slot_object)
```
This is more readable to me. Or you can keep the order you used, but I don't think the two others functions are meaningful.

gui/mozregui/report.py
```diff
@@ -60,11 +61,33 @@ def status_text(self):
...
+    if signal_object and slot_object:
```
parkouss: I would remove the if here. callers have to know how to call this - this is in internal function. Plus I would prefer this to fail instead of silently doing nothing in that case.

gui/mozregui/report.py
```diff
@@ -60,11 +61,33 @@ def status_text(self):
...
+        for name in slots:
+            signal = getattr(signal_object, name)
+            slot = getattr(slot_object, name)
+            if action == "connect":
```
parkouss: you can also use a getattr here:
```
getattr(signal, action)(slot)
```
would do the job in one line. Well maybe a bit less readable.

gui/mozregui/report.py
```diff
@@ -60,11 +61,33 @@ def status_text(self):
...
+                signal.connect(slot)
+            elif action == "disconnect":
+                signal.disconnect(slot)
+            else:
```
parkouss: we don't need this I would say, because it is an internal function.

gui/mozregui/report.py
```diff
@@ -60,11 +61,33 @@ def status_text(self):
...
+def bulk_disconnect_slots(slots, signal_object, slot_object):
+    bulk_action_slots(slots, signal_object, slot_object, 'disconnect')
+
+
 class ReportModel(QAbstractTableModel):
     def __init__(self):
         QAbstractTableModel.__init__(self)
         self.items = []
         self.bisector = None
+        self.is_downloading = False
```
parkouss: hm, could you set this on StepItem instead ? like the progress attribute. This way it may be easy for us when we will implement the background download. For now it should simplify the work of the delegate.

gui/mozregui/report.py
```diff
@@ -73,22 +96,48 @@ def clear(self):
...
+
+    @Slot(object)
+    def download_started(self):
+        self.is_downloading = True
+
+    @Slot(object, str)
+    def download_finished(self, fname):
+        self.is_downloading = False
+
+    def last_item(self):
```
parkouss: not sure if this is useful. If you want to avoid accessing the items variables from outside, maybe just implement an item method. you could make it so that last_item() == item(-1). This would make the function more useful.

gui/mozregui/report.py
```diff
@@ -73,22 +82,53 @@ def clear(self):
 
     @Slot(object)
     def attach_bisector(self, bisector):
-        slots = ('step_started', 'step_build_found', 'step_testing',
-                 'step_finished', 'started', 'finished')
+        bisector_slots = ['step_started',
```
parkouss: Maybe mainly a matter of taste, but I prefer to use tuple instead of list when the data inside won't be changed, because tuple are immutable and they are here for that reason :) Same thing for downloader_slots.

gui/mozregui/report.py
```diff
@@ -73,22 +82,53 @@ def clear(self):
...
+
         self.bisector = bisector
 
+    @Slot(object, int, int)
+    def download_progress(self, dl, current, total):
+        item = self.items[-1]
+        item.progress = (current * 100) / total
+        self.update_item(item)
+
+    def is_downloading(self):
```
parkouss: This should go on the Item class I think (the item hold the downloading attribute).
I think you could do something like:
```
class Item(object):
    def is_downloading(self):
        return False

class StepItem(Item):
    def is_downloading(self):
        return self.downloading
```

gui/mozregui/report.py
```diff
@@ -138,17 +178,20 @@ def step_build_found(self, bisection, build_infos):
             # and add the new step with build_infos
             item = StepItem()
             item.state_text = 'Downloading'
+            item.downloading = True
             item.data.update(build_infos)
             self.append_item(item)
         else:
             # previous item is a step, just update it
             last_item.data.update(build_infos)
+            last_item.downloading = True
```
parkouss: hm, this seems strange. The first upper "item.downloading = True" is not sufficient ?

gui/mozregui/report_delegate.py
parkouss: Here you should just get the item, then paint the download if necessary. In pseudo code:
```
# get the item being painted
item = get_current_painted_item()
if item.is_downloading():
    # paint the download bar
```

gui/mozregui/report.py
```diff
@@ -73,22 +82,53 @@ def clear(self):
...
+        item.progress = (current * 100) / total
+        self.update_item(item)
+
+    def is_downloading(self):
+        item = self.items[-1]
+        if isinstance(item, StepItem):
+            return item.downloading
+        return False
+
+    def get_download_progress(self):
```
parkouss: this should go on the item (or just use item.progress)

PR #232, opened 2015-06-10
Fix Bug 1170093-add a log report view

MikeLing:


gui/mozregui/main.py
```diff
@@ -97,11 +98,33 @@ def show_about(self):
     def edit_global_prefs(self):
         change_prefs_dialog(self)
 
+    @Slot(str)
+    def on_log_received(self, log_message):
+        self.ui.log_information_view.appendPlainText(log_message)
+        counter = self.ui.log_information_view.blockCount()
+        print "counter is ", counter
+        if(counter > 1000):
+            self.ui.log_information_view.clear()
```
parkouss: hmm this will erase everything. I think that we should try to just remove the first block in this case, and after append the log line. Maybe this can help as a start: http://stackoverflow.com/questions/10417795/remove-a-line-block-from-qtextedit

gui/mozregui/main.py
```diff
@@ -97,11 +98,33 @@ def show_about(self):
...
+
+
+class LogModel(QObject):
+    log = Signal(str)
+
+    def __init__(self):
```
parkouss: This is not needed if you just call the base constructor.

gui/mozregui/main.py
```diff
@@ -97,11 +98,33 @@ def show_about(self):
...
+        QObject.__init__(self)
+
+    def __call__(self, data):
+        print "I got a log", data
+        log_message = '{time:d}:{level:s}:{message:s}'.format(**data)
+        print "I got a message", log_message
+        self.log.emit(log_message)
```
parkouss: I would prefer that we send the raw log dict here. Then the QPlainTextEdit will do the formatting.

MikeLing:
I see, but I also notice my pr is failed. It's my first time to see the failure about coverage, could you tell me what should I do to fix it?

Thank you!

parkouss:
Ah, the coverage reflect what part of the code is unit tested. For the gui, we don't have much test code (look in the gui/tests folder). You could add some tests but it is actually pretty hard, so I would say just forget about it for now. :)

Also, I forgot to mention, think about removing your print statements at the end!

MikeLing:
ah, I see. 

Thank you :)

parkouss:
Thanks !
Merged manually with 0b75adee7d920c5bc7befba0cc16e0a2959b0643. I just removed the parenthesis around the if, and added a comment:
https://github.com/mozilla/mozregression/commit/0b75adee7d920c5bc7befba0cc16e0a2959b0643#diff-188bd437ee82445b621418940c03ccafR110

PR #235, opened 2015-06-17
Override default selection theme to keep cells verdict color

jpigree:
Cells Background color represents the verdict on the nightly by the user
of mozregui. But the selection default theme replace it with a blue background.
 
It hides a very useful information for the user.

The fix overrides this behavior and keeps verdict color on selected cells
but bolden its text.


parkouss:
Thanks! Merged manually in ef9f82faa352ab4e9d2373e22bbd651936ae1cbb.

PR #236, opened 2015-06-18
fix resume info escaping

parkouss:


mozregression/bisector.py
```diff
@@ -8,6 +8,7 @@
 import datetime
 import tempfile
 import mozfile
+import pipes
```
wlach: Looks like pipes.quote is deprecated, but there isn't a better option until Python 3.3: https://docs.python.org/2/library/pipes.html#pipes.quote

PR #238, opened 2015-06-21
Fix bad verdict color on selection

jpigree:
For no reason, the QColor for bad verdict also setted alpha-channel
value. When selected, it led to a purple background color.


parkouss:
rebased as c12c0dbe346424bbb4e3ef876880b3d3e7e72bc8.

Thanks!

PR #245, opened 2015-07-02
Bug 1179227-[gui] allow to choose a repository using the wizard

MikeLing:


gui/mozregui/ui/nightlies.ui
```diff
@@ -44,6 +47,26 @@
        </property>
       </widget>
      </item>
+     <item row="2" column="0">
+      <widget class="QLabel" name="label_3">
+       <property name="toolTip">
+        <string>the repository you want to select</string>
```
parkouss: Can you change that with:

you can choose a specific repository name here

gui/mozregui/ui/nightlies.ui
```diff
@@ -44,6 +47,26 @@
...
+       </property>
+      </widget>
+     </item>
+     <item row="2" column="1">
+      <widget class="QLineEdit" name="repository">
+       <property name="toolTip">
+        <string/>
+       </property>
+       <property name="placeholderText">
+        <string> choose a repository you want</string>
```
parkouss: Same thing here:

you can choose a specific repository name here

parkouss:
Yeah, excellent! This works really well, thanks. :) I tried that with the mozilla-aurora repo, this is good!
Please just change the text so the user will understand that this is not required to put something in the text field.

parkouss:
Oh while we are here, could you add in the tooltip the examples we gave in the command line ?
[mozilla-aurora|mozilla-beta|...]
This will help the user to understand what can be put in the field.

Oh I love the text that we can see when there is no value in there!

MikeLing:
Done :)

PR #248, opened 2015-07-03
Taskcluster inbound

parkouss:
Yup, that's it - integration of taskcluster for inbound. :)


mozregression/fetch_configs.py
```diff
@@ -174,47 +173,55 @@ def set_inbound_branch(self, inbound_branch):
...
+    return part
 
-    def inbound_base_urls(self):
-        return [self.root_build_base_url
-                % (self.inbound_branch,
-                   self.build_base_os_part[self.os][self.bits])]
+
+class FirefoxInboundConfigMixin(InboundConfigMixin):
+    def tk_inbound_route(self, changeset):
+        return 'buildbot.revisions.%s.%s.%s' % (
```
wlach: lately I've been more into using format for this sort of thing (i.e. `return 'buildbot.revisions.{}.{}.{}'.format(...)`) as I think it's a little easier to read/understand

mozregression/bisector.py
```diff
@@ -531,13 +531,8 @@ def bisect_nightlies(self, good_date, bad_date):
...
         self._logger.info("Getting inbound builds between %s and %s"
                           % (good_rev, bad_rev))
-        # anything within twelve hours is potentially within the range
-        # (should be a tighter but some older builds have wrong timestamps,
-        # see https://bugzilla.mozilla.org/show_bug.cgi?id=1018907 ...
-        # we can change this at some point in the future, after those builds
-        # expire)
         handler = InboundHandler(find_fix=self.options.find_fix)
-        result = self.do_bisect(handler, good_rev, bad_rev, range=60*60*12)
+        result = self.do_bisect(handler, good_rev, bad_rev)
```
wlach: So are we still downloading a range of a few days before the revision? I really think we need to do that, otherwise we might not be able to isolate the right changeset.

wlach:
Getting the following error when I run it this way:

    (mozregression)Hammersmith:mozregression wlach$ mozregression -g 2015-07-01 -b 2015-07-02
     0:02.23 LOG: MainThread Bisector INFO Got as far as we can go bisecting nightlies...
     0:02.24 LOG: MainThread Bisector INFO Last good revision: 079b6f1ae1c3
     0:02.24 LOG: MainThread Bisector INFO First bad revision: f5e3bacfb60e
     0:02.24 LOG: MainThread Bisector INFO Pushlog:
    https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=079b6f1ae1c3&tochange=f5e3bacfb60e
    
     0:02.24 LOG: MainThread Bisector INFO ... attempting to bisect inbound builds (starting from 4 days ago, to make sure no inbound revision is missed)
     0:03.81 LOG: MainThread Bisector INFO Getting inbound builds between d4c4ce7f060c and f5e3bacfb60e
    Retrieving valid builds from (u'5cde286271f71a8e6b014be0010d57db2d762864', 1435581020) generated an exception: Indexed task not found
 
Doing it for a different date range seems to work fine tho.

wlach:
Also it appears that the code coverage of the unit tests went down, though I don't know if that's really a serious problem.

mozregression/fetch_configs.py
parkouss: seems good to me, will fix it.

mozregression/bisector.py
parkouss: So as discussed on irc, on the ftp the builds were indexed by build timestamp - so we needed that code to find every build folder. This is no more usefull with taskcluster, since it indexes on changesets directly. The code that handle the move from nightly to inbound (going back in time 4 days ago, to not miss changesets) is still there.

parkouss:
In fact there was a bug that you discovered. Sometimes taskcluster does not index task (not sure why, but that happen) and this was not handled. It may be a taskcluster issue, but mozregression should handle that case anyway. It is better to try another changeset than to just stop everything. :) Thanks!

Yeah, the code I just added is not covered by unit test - that's why the global coverage decreased. I'll write a test. (I like to see everything green!)

PR #252, opened 2015-07-17
Bug 1184915 - fix 'InboundBuildData' object has no attribute 'raw_revisions'

parkouss:


mozregression/build_data.py
```diff
@@ -402,6 +404,9 @@ def __init__(self, fetch_config, start_rev, end_rev):
 
         self.info_fetcher = BuildFolderInfoFetcher(None, None)
 
+    def was_empty(self):
```
wlach: Personally I don't see the point of the wrapper method here -- can you not just read/write was_empty directly?

I'd at least make it a property, so you don't have to call the function.

mozregression/bisector.py
```diff
@@ -559,10 +559,11 @@ def bisect_inbound(self, good_rev, bad_rev):
...
             self._logger.info("No inbound data found.")
             # check if we have found revisions
-            if handler.build_data.raw_revisions:
+            if not handler.build_data.was_empty():
                 # if we have, then these builds are probably too old
-                self._logger.info('There are no build dirs on inbound for'
-                                  ' these changesets.')
+                self._logger.info(
+                    'There are no build artifacts on inbound for these'
+                    ' changesets (these are probably too old).')
```
wlach: these are probably -> they are probably

parkouss:
Oh, this was fixed and merged in with 6ddc517f6c36.

PR #256, opened 2015-07-21
Bug 1160078 - With --persist, prioritize already downloaded builds over builds which require a download

parkouss:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1160078.


jpigree:
Okay. I will test it tonight and you will be able to merge it.

PR #231, opened 2015-06-04
Added platform count

joridos:
Referent Bug [1165514 - [gh-pages] show the number of gui releases downloaded](https://bugzilla.mozilla.org/show_bug.cgi?id=1165514)

PR #259, opened 2015-07-23
Bug 1165514 - [gh-pages] on http://mozilla.github.io/mozregression/, …

Gioyik:
…show the number of gui releases downloaded

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -18,4 +18,24 @@ layout: default
       </li>
     {% endfor %}
   </ul>
+
+  <br>
+  <h3>Number of downloads for the gui version:</h3>
+  <div id="downloads"></div>
+  <script>
+    function createElement(version1, version2) {
+      var tag = document.createElement('dl');
+      var template = '<dl>'
+      +  '<dd>Windows: ' + version1 + '</dd>'
+      +  '<dd>Linux: ' + version2 + '</dd>';
+      tag.innerHTML = template;
+      document.getElementById('downloads').appendChild(tag);
+    }
+    var tags = {};
+    $.ajax({
+      url: "https://api.github.com/repos/mozilla/mozregression/releases"
+    }).done(function(data1) {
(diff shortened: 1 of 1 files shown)


_layouts/home.html
```diff
@@ -18,4 +18,24 @@
       </li>
     {% endfor %}
   </ul>
+
+  <br>
+  <h3>Number of downloads for the gui version:</h3>
+  <div id="downloads"></div>
+  <script>
+    function createElement(version1, version2) {
```
parkouss: Could you rename version1 and version2 to something like win_version and linux_version ?

_layouts/home.html
```diff
@@ -18,4 +18,24 @@
...
+      +  '<dd>Windows: ' + version1 + '</dd>'
+      +  '<dd>Linux: ' + version2 + '</dd>';
+      tag.innerHTML = template;
+      document.getElementById('downloads').appendChild(tag);
+    }
+    var tags = {};
+    $.ajax({
+      url: "https://api.github.com/repos/mozilla/mozregression/releases"
+    }).done(function(data1) {
+      createElement(data1[0].assets[0].download_count, data1[0].assets[1].download_count);
```
parkouss: This is only the number of download for the latest release - and I'm not sure that 0 will always be windows (and 1 linux), I suspect it depends on the time required to build the release (we are using travis-ci and appveyor for that).

Could you make the change to count the number of downloads for all release ? Also I would prefer to look at the "name" property of an asset (instead of using indexes): if it ends with ".exe", this is for windows, else if it ends with ".tar.gz" this is for linux (please do the two tests as we will add mac one day or the other).

parkouss:
Looks pretty good, thanks @Gioyik ! I noted some comments, could you address them please ?

parkouss:
Closed as duplicate of #231. Thanks @Gioyik for the initial work. :)

PR #262, opened 2015-07-24
Bug 1187302  [gui] python 'gui/build.py run' fails on mac

MikeLing:


gui/build.py
```diff
@@ -60,7 +60,13 @@ def do_run(options):
     do_uic(options)
     do_rcc(options)
     env = dict(os.environ)
-    env['PYTHONPATH'] = '.'
+    print sys.platform
+    # if runing under the OS X, we should just add '.' to
+    # our PATHONPATH rather than override it
+    if env['PYTHONPATH'] != '':
```
parkouss: hm, this should be "if 'PYTHONPATH' in env"

gui/build.py
```diff
@@ -60,7 +60,13 @@ def do_run(options):
     do_uic(options)
     do_rcc(options)
     env = dict(os.environ)
-    env['PYTHONPATH'] = '.'
+    print sys.platform
```
parkouss: could you remove that print ?

gui/build.py
```diff
@@ -60,7 +60,13 @@ def do_run(options):
...
+    # if runing under the OS X, we should just add '.' to
```
parkouss: the comments needs to be updated now. maybe just:
# update PYTHONPATH so python can find mozregui package

gui/build.py
```diff
@@ -60,7 +60,11 @@ def do_run(options):
     do_uic(options)
     do_rcc(options)
     env = dict(os.environ)
-    env['PYTHONPATH'] = '.'
+    # update PYTHONPATH so python can find mozregui package
+    if env.get('PYTHONPATH') != '':
```
parkouss: This line fails with an error for me:
ERROR: 'PYTHONPATH'

You should just write
```
if env.get('PYTHONPATH'):
```
This is because the get() method of a dict returns None by default when the key does not exist. and None != ''.
https://docs.python.org/2/library/stdtypes.html#dict.get

PR #264, opened 2015-07-30
add post for the 0.40 release

parkouss:


_posts/2015-07-30-0.40-release.md
```diff
@@ -0,0 +1,14 @@
+---
+title:  "Release 0.40"
+author: Julien Pagès
+---
+
+This is it, now mozregression uses fully the new build storage!
+
+From what I tested so far, this is a lot faster to download builds
+from Europe. Good news!
```
wlach: I think lines 6-9 might be confusing to people who are not familiar with the ftp.mozilla.org migration. I'd just say something like: `mozregression now downloads builds from archive.mozilla.org. In many cases this should be faster than before. Good news!"

PR #266, opened 2015-08-05
Clone existing mozprofiles to prevent persistence

jpigree:
* When using an existing mozprofile, mozregression modifies it without
reverting the changes after the tests. This fix makes it clone the
existing profile in order to preserve its state.

* This fix needs mozprofile >= 0.24 in order to clean up the profile
created into the tmp directory when no reference to the object is left.

=> Fortunately, setup.py is already updated to 0.25 so no change needed.

* Adapted launcher unit tests accordingly.

* Factorized multiple hard coded strings into launcher test cases.

### DIFF
--- a/tests/unit/test_launchers.py
+++ b/tests/unit/test_launchers.py
@@ -22,6 +22,7 @@ class MyLauncher(launchers.Launcher):
 
 
 class TestLauncher(unittest.TestCase):
+
     def test_start_stop(self):
         launcher = MyLauncher('/foo/persist.zip')
         self.assertFalse(launcher.started)
@@ -39,6 +40,7 @@ class TestLauncher(unittest.TestCase):
 
 
 class TestMozRunnerLauncher(unittest.TestCase):
+
     @patch('mozregression.launchers.mozinstall')
     def setUp(self, mozinstall):
         mozinstall.get_binary.return_value = '/binary'
@@ -81,9 +83,8 @@ profile_class', spec=Profile)
     def test_start_with_profile_and_addons(self, Runner):
         self.launcher_start(profile='my-profile', addons=['my-addon'],
                             preferences='my-prefs')
-        self.profile_class.assert_called_once_with(profile='my-profile',
-                                                   addons=['my-addon'],
-                                                   preferences='my-prefs')
+        self.profile_class.clone.assert_called_once_with(
+            'my-profile', addons=['my-addon'], preferences='my-prefs')
         # runner is started
         self.launcher.runner.start.assert_called_once_with()
         self.launcher.stop()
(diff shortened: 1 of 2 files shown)


tests/unit/test_launchers.py
```diff
@@ -105,10 +106,22 @@ def test_launcher_deleted_remove_tempdir(self):
 
 
 class TestFennecLauncher(unittest.TestCase):
+
+    def setUp(self):
+        self.test_root = '/sdcard/tmp'
+        self.profile = Profile()
+        self.profile_name = os.path.basename(self.profile.profile)
+        # No join because there is a path check later on
```
parkouss: On remote devices, the separator is always '/' - so the line after is normal. Can you remove or update the comment ?

tests/unit/test_launchers.py
```diff
@@ -105,10 +106,22 @@ def test_launcher_deleted_remove_tempdir(self):
 
 
 class TestFennecLauncher(unittest.TestCase):
+
+    def setUp(self):
+        self.test_root = '/sdcard/tmp'
+        self.profile = Profile()
+        self.profile_name = os.path.basename(self.profile.profile)
```
parkouss: It seems that this variable is only used after - could you not attach it to self - or even better do not use a variable ?

tests/unit/test_launchers.py
```diff
@@ -105,10 +106,22 @@ def test_launcher_deleted_remove_tempdir(self):
...
+        # No join because there is a path check later on
+        self.remote_profile_path = self.test_root + '/' + self.profile_name
+
+    def tearDown(self):
+        # explicit cleanup to force profile directory deletion
+        self.profile.cleanup()
```
parkouss: I would prefer to see this in the setUp method, registered with addCleanup: https://docs.python.org/2/library/unittest.html#unittest.TestCase.addCleanup
This is less verbose and will be called even if the test fail.

tests/unit/test_launchers.py
jpigree: It is used twice. I created a class member for it.

tests/unit/test_launchers.py
```diff
@@ -105,10 +106,19 @@ def test_launcher_deleted_remove_tempdir(self):
 
 
 class TestFennecLauncher(unittest.TestCase):
+
+    test_root = '/sdcard/tmp'
+
+    def setUp(self):
+        self.profile = Profile()
+        self.addCleanup(self.profile.cleanup)
+        self.profile_name = os.path.basename(self.profile.profile)
```
parkouss: Oh I meant the "profile_name" variable, not test_root. :)

jpigree: Urgh. Sorry. But the test_root needed to be a class member instead of reaffected on each test run with setUp. I fix profile_name aswell.

PR #267, opened 2015-08-05
1191254 empty directories

jpigree:
Okay. As I said in the issue, it seems that py.tests does not execute the __del__ methods. So I add to explicitly call them (addCleanUp).

### DIFF
--- a/tests/unit/test_launchers.py
+++ b/tests/unit/test_launchers.py
@@ -55,6 +55,7 @@ profile_class', spec=Profile)
 
     def test_installed(self):
         self.assertEqual(self.launcher.binary, '/binary')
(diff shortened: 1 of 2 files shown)


tests/unit/test_launchers.py
```diff
@@ -55,6 +55,7 @@ def launcher_start(self, profile_class, *args, **kwargs):
 
     def test_installed(self):
         self.assertEqual(self.launcher.binary, '/binary')
+        del self.launcher
```
jpigree: I had to delete manually on each functions of this test case because the last test case tests that a call to __del__ actually delete the tempdir.

PR #270, opened 2015-08-11
Bug 1192487 - change the nightlies data format to yyyy-mm-dd

MikeLing:


gui/mozregui/ui/nightlies.ui
```diff
@@ -29,7 +29,7 @@
      <item row="0" column="1">
       <widget class="QDateEdit" name="start_date">
        <property name="displayFormat">
-        <string>dd/MM/yyyy</string>
+        <string>yyyy-mm-dd</string>
```
parkouss: hmm, I think it should be yyyy-MM-dd, because mm stands for minute, not months. :)

MikeLing: Oh, sorry about that. My mistake :(

PR #273, opened 2015-08-13
Bug 1192850 - Add Arch/Manjaro specific install steps

jpigree:


install.md
```diff
@@ -49,6 +49,10 @@ follow the OS-specific instructions to install pip if you don't know how to do i
 
       sudo apt-get install python-pip
 
+- **Arch / Manjaro (Python3 distributions)**:
```
parkouss: Hmm I think it would be better to add note (maybe inside the ubuntu block - or a new one like you did) for linux users, related to all distros that use python3 by default. The example is good though, and it is good to have it, but I would add a more general comment so we won't add a block for every distro.

parkouss: Or just changing the title, instead of "Arch / Manjaro (Python3 distributions)", something like "Python 3 linux distributions (like archlinux)".

jpigree:
Okay. I added a note for Python3 distros.

jpigree:
Okay. Is it better?

parkouss:
Hmm, looking at this, I don't like the "if default" inside Ubuntu sections since ubuntu is a python2 only (no choice for users, this can be confusing).
What about a new section like:
```
**Other linuxes**
  be careful that some distributions uses python 3 by default, in this case you need to install *python2-pip*.
  Example for archlinux:
    sudo pacman -S python2-pip
```
And I would recommend to use *pip2 install* for Ubuntu and other linuxes, since it should work everywhere. (Just add a section "Linux" using pip2 and make a mac only section for using pip since I am not sure about mac).

Is it clear ?

Yeah, this is harder than I thought.: ) Thanks for looking at it. :)

jpigree:
Yep. Did not found how to write this. I made another try. Feel free to review it. Thanks.

parkouss:
Perfectly fine for me. Thanks!

PR #275, opened 2015-08-14
Bug 1194219 - [gui] add the ability to use addons 

MikeLing:


gui/mozregui/addons_editor.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def flags(self, index):
+        return Qt.ItemIsSelectable | Qt.ItemIsEditable | Qt.ItemIsEnabled
+
+    def add_addons_from_file(self, addons):
+        if addons:
+            path = len(self.addonsPath)
+            self.beginInsertRows(QModelIndex(), path,
+                                 path + len(addons) - 1)
+            self.addonsPath.extend(str(addons))
```
MikeLing: After I insert some path like "/User/mikeling/mozregression/README.md", it will become something like ['/', 'U', 's', 'e', 'r', 's', '/', 'm', 'i', 'k', 'e', 'l', 'i', 'n', 'g', '/', 'm', 'o', 'z', 'r', 'e', 'g', 'r', 'e', 's', 's', 'i', 'o', 'n', '/', 'R', 'E', 'A', 'D', 'M', 'E', '.', 'm', 'd'] in self.addonsPath. And I don't know where goes wrong. 

I guess it's wrong because I define a wrong setData method. But, actually, I haven't fully understand setData method even after read the document about it. Could you help me about that? Thank you very much! @parkouss

parkouss: I think you just need to use 'append' instead of 'extend'. But I think you misunderstood what a addon is. An addon for firefox is an extension (I think you new that but let's be clear), https://addons.mozilla.org/en-US/firefox/. Here we should give a list of addons, so I would say a list of file end each file name ends with '*.xpi'.

I think that the Profile class we use can also take a folder name that contains addons - but I am not sure, and this won't do us any good from the gui. I think we should just allow the user to select one or more *.xpi files.

Does that help ?

gui/mozregui/addons_editor.py
```diff
@@ -0,0 +1,85 @@
...
+        self.addonsPath = []
+
+    def rowCount(self, index=QModelIndex()):
+        return len(self.addonsPath)
+
+    def data(self, index, role=Qt.DisplayRole):
+        if role == Qt.DisplayRole:
+            return self.addonsPath[index.row()]
+
+    def headerData(self, section, orientation, role=Qt.DisplayRole):
```
parkouss: you can remove this function, since you don't use the table headers.

gui/mozregui/addons_editor.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class AddonsMode(QAbstractListModel):
+    """
+    A Qt model that can edit addons path.
+    """
+
+    def __init__(self, parent=None):
+        QAbstractListModel.__init__(self, parent)
+        self.addonsPath = []
```
parkouss: please just name this "addons"

gui/mozregui/addons_editor.py
```diff
@@ -0,0 +1,85 @@
...
+            return self.addonsPath[index.row()]
+
+    def headerData(self, section, orientation, role=Qt.DisplayRole):
+        if role == Qt.DisplayRole and orientation == Qt.Horizontal:
+            return 'path'[section]
+
+    def flags(self, index):
+        return Qt.ItemIsSelectable | Qt.ItemIsEditable | Qt.ItemIsEnabled
+
+    def add_addons_from_file(self, addons):
```
parkouss: you shoud rename this function with add_addon(self, addon)

gui/mozregui/addons_editor.py
```diff
@@ -0,0 +1,85 @@
...
+    def __init__(self, parent=None):
+        QWidget.__init__(self, parent)
+        self.ui = Ui_AddonsEditor()
+        self.ui.setupUi(self)
+
+        self.list_model = AddonsMode()
+        self.ui.list_view.setModel(self.list_model)
+
+    @Slot()
+    def add_addons_from_file(self):
```
parkouss: rename to add_addons()

gui/mozregui/addons_editor.py
```diff
@@ -0,0 +1,85 @@
...
+        fname = QFileDialog.getOpenFileName(
```
parkouss: you could use http://doc.qt.io/qt-4.8/qfiledialog.html#getOpenFileNames, so the user can choose multiple addons at the same time

gui/mozregui/addons_editor.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def flags(self, index):
+        return Qt.ItemIsSelectable | Qt.ItemIsEditable | Qt.ItemIsEnabled
+
+    def add_addons_from_file(self, addons):
+        if addons:
+            addons_list_length = len(self.addonsPath)
+            self.beginInsertRows(QModelIndex(), addons_list_length,
+                                 addons_list_length)
+            self.addonsPath.append(str(addons))
```
parkouss: remove the str() here - else it may break for some paths that are non unicode

gui/mozregui/bisection.py
```diff
@@ -256,6 +256,10 @@ def bisect(self, fetch_config, options):
                 value = options[name]
                 if value:
                     launcher_kwargs[name] = value
+
+        # add add-ons path to the app luncher
+        for path in options['addons']:
+            launcher_kwargs['addons'] = path
```
parkouss: this was the error, just write "auncher_kwargs['addons'] = options['addons']"

parkouss:
I noted some comments. Overall, this is great!

gui/mozregui/addons_editor.py
```diff
@@ -0,0 +1,90 @@
+from PyQt4.QtCore import QAbstractListModel, QString, QModelIndex, Qt, \
+    pyqtSlot as Slot
+from PyQt4.QtGui import QWidget, QFileDialog
+from mozregui.ui.addons_editor import Ui_AddonsEditor
+
+
+class AddonsMode(QAbstractListModel):
```
parkouss: Typo: AddonsModel

gui/mozregui/addons_editor.py
```diff
@@ -0,0 +1,90 @@
+from PyQt4.QtCore import QAbstractListModel, QString, QModelIndex, Qt, \
```
parkouss: It seems that QString is not used - this make the travis-ci/appveyor builds unhappy, see details from https://github.com/mozilla/mozregression/pull/275 (https://travis-ci.org/mozilla/mozregression/builds/76215027#L492)

gui/mozregui/addons_editor.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def rowCount(self, index=QModelIndex()):
+        return len(self.addons)
+
+    def data(self, index, role=Qt.DisplayRole):
+        if role == Qt.DisplayRole:
+            return self.addons[index.row()]
+
+    def flags(self, index):
+        return Qt.ItemIsSelectable | Qt.ItemIsEditable | Qt.ItemIsEnabled
```
parkouss: Please remove Qt.ItemIsEditable, This is not used in this model

gui/mozregui/bisection.py
```diff
@@ -256,6 +256,9 @@ def bisect(self, fetch_config, options):
                 value = options[name]
                 if value:
                     launcher_kwargs[name] = value
+
+        # add add-ons path to the app luncher
```
parkouss: typo:
add add-ons path**s** to the app l**a**uncher

gui/mozregui/addons_editor.py
```diff
@@ -0,0 +1,90 @@
...
+    @Slot()
+    def clear_addons(self):
+        self.list_model.clear_addons()
+        self.ui.list_view.reset()
+
+    def get_addons(self):
+        return self.list_model.addons
+
+
+if __name__ == '__main__':
```
parkouss: Later (in another bug), we could add a real unit test to test this code.

parkouss:
Thanks for the update!

Merged manually in bd1110e2.

PR #283, opened 2015-08-19
1195082 main split

jpigree:
- Created cli module and its associated test_cli.py unittests
- Move args parser from main to cli module
- Moved args parser (which now concerns only defaults handling) tests from  test_main.py to test_cli.py
- Fix preexisting typo

### DIFF
--- a/mozregression/cli.py
+++ b/mozregression/cli.py
@@ -1,0 +1,182 @@
+#!/usr/bin/env python
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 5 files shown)


jpigree:
Okay. I splitted the main.py as requested. I moved the argument parser into the cli.py. Seems pretty good, except the method cli which is actually into the main.py but it is not args parsing so I can't move it into cli.py. Perhaps should I rename it run?

I moved the tests aswell. The test_cli.py perhaps needs some more tests but they are already done through the main.py, so I think it is enough.

Feel free to comment this.

jpigree:
Okay. My questions were already answered in Bug 1195195. Great!!!

PR #289, opened 2015-08-23
Bug 1194937 - mozregression default 'good' date 2009-01-01 doesn't work on win64

parkouss:


parkouss:
Landed manually in https://github.com/mozilla/mozregression/commit/570a9fb293295a1520b5c578e4b860e99f186f55.

PR #290, opened 2015-08-24
Bug 1197744 - add a TestRunner.run_once method

parkouss:


mozregression/launchers.py
```diff
@@ -216,6 +233,10 @@ def _start(self, profile=None, addons=(), cmdargs=(), preferences=None):
         self.adb.launch_fennec(self.package_name,
                                extra_args=["-profile", self.remote_profile])
 
+    def _wait(self):
+        while self.adb.process_exist():
+            time.sleep(1)
```
wlach: I'd use a shorter timeout than this.

tests/unit/test_launchers.py
```diff
@@ -63,11 +76,15 @@ def test_stop_fail(self):
 
 
 class TestMozRunnerLauncher(unittest.TestCase):
+    def clean_launcher(self):
+        if hasattr(self, 'launcher'):
+            del self.launcher
```
wlach: I generally don't really like relying on destructors in python for behavior, as it's kind of unintuitive. If there's cleanup code that needs to be run when we're finished, I'd rather have a cleanup() method that's called explicitly.

parkouss: Totally agree. but it is here for quite a long time - good news is that it is the only last thing in mozreg that rely on destructors. A bug to fix that would be great, I'll create that.

mozregression/launchers.py
parkouss: Make sense to me. hmm, 0.3 ?

parkouss:
Landed manually in https://github.com/mozilla/mozregression/commit/74322ef648df22a250bc5f7a321fe790f828a6db

PR #292, opened 2015-08-26
Bug 1198866 - help the user to use a config file

parkouss:


jpigree:
Seems obvious but the appveyor failed on your new tests.

jpigree:
Hum. Your commit is way less verbose than I am accustomed with.

jpigree:
Hum Seems good. The color with blessings helps find quickly the value we choose.

jpigree:
If I have to be annoying, I would say that it is strange to have a blank line before any outputs when using --write-config

jpigree:
Strange. I just did the --write-config. I used all default values but when looking at the config file generated, it lacks bits=64???

jpigree:
Is it what we want?

parkouss:
Hum, sorry for fixing my PR by rebasing wile you was adding comments - it seems taht github does not support it very well. And now I have at least some of your messages but withouth context... (And I just added parenthesis for one line!)

Well, I do like the empty line! Let's see with a third person. :)

the bits = 64 is not here in the config file yes, and this is normal (because 64 bits is the default in command line). This was intentional, but now that I think about it, this is not ideal. (Better write the value, so it would be easy for a user to change it).

Thanks a lot for the review @jpigree ! I have a last question: We had a discussion with @wlach, and I think he was more in favor for something that ask directly the user (ie no need for --write-conf). This is another point of view, but this may be blocking for the user trying to use the --command for example so we could have to add a --non-interactive flag or something.

Anyway - do you have a preference for this ? If you were a user how would you like this to be ?

parkouss:
Ok, I just updated the code to make some changes:

- no more sys.exit from config module (raising an exception and handling it in main - I agree this is cleaner!)
- always write values in the generated config files

jpigree:
Okay. Great. I only have a question left.
Why try, catch on blessings when it is clearly a dependency in your setup.py?

jpigree:
For the choice on --write-config, I like the git interface for config files.
git config -e => open the configuration file
git config --list => list all keys

When key elements lacks, it throws warnings and even prevent you from committing (email/name).
You have to use the config command.

parkouss:
Well, blessings on windows requires curses, that can not be required by pip. So I need to try/catch because I don't want to force users to install that, but I would like some color if available (mozlog does the same thing).

For the other point, if the program throw an error and exit for some configuration that is not required, then it is not fully non interactive I think: it can fail after an upgrade, just because we add an option that we would like the user to think of. So it becomes hard to have a script that call mozregression and is really not interactive, thus we should in that case add a --no-interactive flag or such to e.g not check configuration (just bisect).

PR #293, opened 2015-08-28
Bug 1199618 - mozregression 0.41 fails to find inbound builds for revisions from this month

parkouss:


mozregression/fetch_build_info.py
```diff
@@ -88,8 +88,21 @@ def find_build_info(self, changeset, fetch_txt_info=True):
         try:
             task_id = self.index.findTask(tk_route)['taskId']
         except TaskclusterFailure:
-            raise BuildInfoNotFound("Unable to find build info using the"
-                                    " taskcluster route %r" % tk_route)
+            # HACK because of
+            # https://bugzilla.mozilla.org/show_bug.cgi?id=1199618
+            # TODO remove the if once these inbounds no more exists
```
wlach: TODO remove the if once these inbounds no more exists (just raise our BuildInfoNotFound) -> TODO: remove the if statement once these builds no longer exist (just raise BuildInfoNotFound)

PR #299, opened 2015-09-01
Bug 1147576 - add some documentation for the special case of float pref

parkouss:


parkouss:
Landed manually in https://github.com/mozilla/mozregression/commit/efa584fff699d96248e563582783dac974052523.

PR #304, opened 2015-09-04
Bug 1194941 - Offer a way to launch a specific build

MikeLing:


mozregression/cli.py
```diff
@@ -389,9 +392,16 @@ def validate(self):
             # this can be useful for both inbound and nightly, because we
             # can go to inbound from nightly.
             fetch_config.set_inbound_branch(options.inbound_branch)
+        # set action for just use changset or data to bisect
+        if options.launch:
+            try:
+                options.launch = parse_date(options.launch)
```
MikeLing: Hmm, Some errors shows up when I use --launch 2014-12-15
1:35.07 LOG: MainThread ERROR HTTPSConnectionPool(host='archive.mozilla.org', port=443): Max retries exceeded with url: /pub/mozilla.org/firefox/nightly/2014/12/ (Caused by ConnectTimeoutError(<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x1036d9990>, 'Connection to archive.mozilla.org timed out. (connect timeout=30.0)'))

And I print the options.launch, it still shows '2014-12-15' rather than date(2014,12,15) which it may should be.

parkouss:
This works really good for me! @MikeLing, please rebase on top of master, maybe this will make it work for you too (tell me).

Awesome. :)

parkouss:
The option launch seems correct (it is normal that a date is printed as 2014-12-15), I just thing you had an issue with the network (maybe archive.mozilla.org was down).

parkouss:
Thanks! fixed a couple of nits and landed in https://github.com/mozilla/mozregression/commit/215a875e3c7305c026c0cf25d5fb863d016c15bb

PR #307, opened 2015-09-07
Bug 1202238 - [gui] Mozregression always change app item if switch bits options

MikeLing:


parkouss:
I would prefer to use "self.ui.app_combo.currentText()" instead of "self.ui.app_combo.currentIndex()" so we can avoid the "app_combo_index" variable.

Also could you change your commit message to "Bug 1202238 - [gui] Mozregression always downloads "b2g" if I choose "32 bit" ? I think it more understandable.

MikeLing:
Sure! No problem :)

parkouss:
Cool, just fixed the commit message, and landed in https://github.com/mozilla/mozregression/commit/c6d619ccb052f56b20d51a79d07daa6c5b599839.

PR #309, opened 2015-09-08
Bug 1202203 - [GUI] Make GUI use the same config file as the cli version

MikeLing:


gui/mozregui/global_prefs.py
```diff
@@ -1,17 +1,17 @@
...
 from mozregression.network import set_http_session
-
+from mozregression.config import DEFAULT_CONF_FNAME, get_defaults
 
 def get_prefs():
     """
     Return the global prefs as a dict.
     """
-    settings = QSettings()
+    settings = get_defaults(DEFAULT_CONF_FNAME)
```
parkouss: So yeah, this is the right thing to do. But now the settings object is a dict, not anymore a QSettings, so you have to change the next lines.

gui/mozregui/global_prefs.py
```diff
@@ -23,7 +23,7 @@ def get_prefs():
 
 
 def save_prefs(options):
-    settings = QSettings()
+    settings = get_defaults(DEFAULT_CONF_FNAME)
     settings.setValue("globalPrefs/persist",
```
parkouss: Weel, here you have to use configobj to save the new values.

gui/mozregui/main.py
```diff
@@ -76,12 +76,12 @@ def clear(self):
         mozfile.remove(self.persist)
 
     def read_settings(self):
-        settings = QSettings()
+        settings = get_defaults(DEFAULT_CONF_FNAME)
```
parkouss: These options are GUI only - just keep the QSettings here.

gui/mozregui/main.py
```diff
@@ -76,12 +76,12 @@ def clear(self):
...
         self.restoreGeometry(settings.value("mainWin/geometry").toByteArray())
         self.restoreState(settings.value("mainWin/windowState").toByteArray())
 
     def closeEvent(self, evt):
-        settings = QSettings()
+        settings = get_defaults(DEFAULT_CONF_FNAME)
```
parkouss: These options are GUI only - just keep the QSettings here.

gui/mozregui/global_prefs.py
```diff
@@ -1,33 +1,34 @@
...
     Return the global prefs as a dict.
     """
-    settings = QSettings()
+    settings = get_defaults(DEFAULT_CONF_FNAME)
     options = {}
     options['persist'] = \
-        str(settings.value("globalPrefs/persist",
-                           QVariant("")).toString()) or None
+        str(settings.get("globalPrefs/persist",
+                         "").toString()) or None
```
parkouss: So, you should remove the "globalPrefs/" part of the option name, now that we use the config file.

Also toString() is for QString, but now we are only using python str. The code should look like:

options['persist'] = setting['persist'] or None
options['http_timeout'] = settings['http-timeout']

gui/mozregui/global_prefs.py
```diff
@@ -1,33 +1,34 @@
...
-    settings = QSettings()
-    settings.setValue("globalPrefs/persist",
-                      QVariant(options['persist'] or ''))
-    settings.setValue("globalPrefs/http_timeout",
-                      QVariant(options['http_timeout']))
+    settings = get_defaults(DEFAULT_CONF_FNAME)
+    settings.update(ConfigObj("globalPrefs/persist",
+                              (options['persist'] or '')))
+    settings.update(ConfigObj("globalPrefs/http_timeout",
+                              (options['http_timeout'])))
```
parkouss: Well save_prefs should write the new values, but here you are just updating a dict that is lost just after.

So, I think this should be more like:

settings = ConfigObj(DEFAULT_CONF_FNAME)

then update the settings, and at the end write to a file. This have to be tested, using the Global preferences dialog.

MikeLing:
ok, I think it's good to go now ;) @parkouss

gui/mozregui/global_prefs.py
```diff
@@ -1,33 +1,31 @@
...
 
 def save_prefs(options):
-    settings = QSettings()
-    settings.setValue("globalPrefs/persist",
-                      QVariant(options['persist'] or ''))
-    settings.setValue("globalPrefs/http_timeout",
-                      QVariant(options['http_timeout']))
+    settings = ConfigObj(DEFAULT_CONF_FNAME)
+    print settings
+    print options
```
parkouss: Please remove those prints. :)

gui/mozregui/global_prefs.py
```diff
@@ -1,33 +1,31 @@
...
+    settings.update(ConfigObj({'persist': options['persist'] or ''}))
+    settings.update(ConfigObj({'http-timeout': options['http_timeout']}))
```
parkouss: Well you can remove ConfigObj I think, i.e:

```
settings.update({
    'persist': options['persist'] or '',
    'http-timeout': options['http_timeout'],
})
```

PR #312, opened 2015-09-11
Bug 1202494 - [gui] add the golbal persist size limit option

MikeLing:


gui/mozregui/global_prefs.py
```diff
@@ -16,6 +16,7 @@ def get_prefs():
     options = dict()
     options['persist'] = settings['persist']
     options['http_timeout'] = float(settings['http-timeout'])
+    options['persist-size-limit'] = settings['persist-size-limit']
```
parkouss: This should be:

```
options['persist_size_limit'] = float(settings['persist-size-limit'])
```
(note the use of underscores instead of dashes). We use dashes in the conf file, but underscores elsewhere.

Also note that we need to convert it to float.

gui/mozregui/global_prefs.py
```diff
@@ -24,6 +25,8 @@ def save_prefs(options):
     settings.update({
         'persist': options['persist'] or '',
         'http-timeout': options['http_timeout'],
+        'persist-size-limit': int(abs(float(options.persist_size_limit))
```
parkouss: Looks good, but *options.persist_size_limit* won't work! You should use options['persist_size_limit'].
Also the float conversion is not required here I think since we did that in get_prefs().

gui/mozregui/global_prefs.py
```diff
@@ -48,8 +51,10 @@ def change_prefs_dialog(parent=None):
     options = get_prefs()
     ui.persist.line_edit.setText(options['persist'] or '')
     ui.http_timeout.setValue(options['http_timeout'])
+    ui.persist_size_limit.setValue(options['persist-size-limit'])
```
parkouss: Should be:
```
ui.persist_size_limit.setValue(options['persist_size_limit'])
```

parkouss:
This is good thanks! I added some comments, and it does not work as it is but overall this is great, you are looking into the right direction. +1

gui/mozregui/global_prefs.py
parkouss: hum, thinking more about it, I think we should do that conversion from GiB to bytes in bisection.py, along with:

```
persist_limit = global_prefs['persist_size_limit']
```

gui/mozregui/ui/global_prefs.ui
```diff
@@ -33,6 +33,23 @@
       <widget class="DirectorySelectWidget" name="persist" native="true"/>
      </item>
      <item row="1" column="0">
+         <widget class="QLabel" name="label">
+             <property name="toolTip">
+                 <string>The limit size of persist directory.</string>
```
parkouss: "Size limit of the persist directory, in GiB." would be more appropriate here.

gui/mozregui/ui/global_prefs.ui
```diff
@@ -33,6 +33,23 @@
...
+             </property>
+             <property name="text">
+                 <string>persist size limit:</string>
+             </property>
+         </widget>
+     </item>
+     <item row="1" column="1">
+         <widget class="QDoubleSpinBox" name="persist_size_limit">
+             <property name="minimum">
+                 <double>1.0000</double>
```
parkouss: Please put the minimum to 0.0 (it would mean deactivated), and the max to more than 99.99 (999.99 should be good).

MikeLing:
Done ;) @parkouss

PR #314, opened 2015-09-17
Bug 1205560 - B2G device manual regression finding

mshal:


mozregression/cli.py
```diff
@@ -118,6 +118,14 @@ def create_parser(defaults):
                         help=("last known good revision (for inbound"
                               " bisection)."))
 
+    parser.add_argument("--taskcluster",
+                        help=("the Taskcluster build product/name/type,"
+                              "such as 'b2g.aries-opt'"))
```
parkouss: nit, missing space before such.

mozregression/cli.py
```diff
@@ -353,6 +361,26 @@ def check_inbounds(options, fetch_config, logger):
...
 
+def check_taskcluster(options, fetch_config, logger):
+    if not options.artifact_name:
+        raise MozRegressionError('--artifact-name is required for taskcluster'
+                                 ' regression finding')
+    try:
+        # Try an inbound-style revision range first
+        check_inbounds(options, fetch_config, logger)
+        print "Using revs: ", options.last_good_revision, options.first_bad_revision
+    except:
```
parkouss: Hm, I am not really a fan of except without exception type, since this can catch KeyboardInterrupt and such. Maybe replace the try with:
```
if options.last_good_revision and options.first_bad_revision:
    ...
```

mozregression/cli.py
```diff
@@ -390,6 +418,7 @@ def validate(self):
         options.bits = parse_bits(options.bits or mozinfo.bits)
         fetch_config = create_config(options.app, mozinfo.os, options.bits)
         self.fetch_config = fetch_config
+        self.fetch_config.artifact_name = options.artifact_name
```
parkouss: hm, could you move that next to fetch_config.set_build_type(options.taskcluster) in the special case taskcluster handling ?
also a fetch_config.set_artifact_name would make sense, I mean we will set the two options in the same way, and at the same place.

mozregression/fetch_configs.py
```diff
@@ -68,6 +68,7 @@ class CommonConfig(object):
     """
 
     app_name = None
+    artifact_name = None
```
parkouss: I would prefer to see this in B2GDeviceConfigMixin I think (same level as inbound_branch=None).

mozregression/fetch_configs.py
```diff
@@ -269,6 +270,24 @@ def tk_inbound_route(self, changeset):
...
+
+    def tk_inbound_route(self, changeset):
+        print "Querying inbound route: 'gecko.v2.{}.revision.{}.{}'".format(
+            self.inbound_branch, changeset, self.build_type,
+        )
+        return 'gecko.v2.{}.revision.{}.{}'.format(
+            self.inbound_branch, changeset, self.build_type,
+        )
+
+    def set_nightly_repo(self, repo):
```
parkouss: hm, I suspect you had to do that because we never had a case with an inbound only mixin. We should fix that in a followup, but for now can you move that method into B2GDeviceConfig (since this is for nightly, it would make more sense)

mozregression/json_pushes.py
```diff
@@ -76,3 +81,24 @@ def pushlog_within_changes(self, fromchange, tochange):
         ))
         # sort pushlogs by date
         return sorted(chsets.itervalues(), key=lambda push: push['date'])
+
+    def revision_for_date(self, date, last=False):
+        """
+        Returns the json pushlog object that matches the given date.
```
parkouss: nit, it returns a changeset, not a pushlog object I think

PR #318, opened 2015-09-26
Bug 1197776 - mozregression should warn against using --bits=32 on mac

kapy2010:


mozregression/cli.py
```diff
@@ -442,6 +442,10 @@ def validate(self):
             # inform users on windows that we are using 64 bit builds.
             self.logger.info("bits option not specified, using 64-bit builds.")
 
+        if options.bits == 32 and mozinfo.os == 'mac':
+            self.logger.info("only 64-bit builds available for mac, using \
+64-bit builds")
```
parkouss: Usually in mozregression we split long text lines the following way:
```
    self.logger.info("only 64-bit builds available for mac, using"
                             " 64-bit builds")
```
Note that the quotes are aligned, and there is a space at the beginning of the second line. Could you change that please ?

Thanks!

PR #319, opened 2015-09-26
allow to bisect non integration branches (like m-c) on b2g-devices

parkouss:


mozregression/fetch_configs.py
```diff
@@ -222,11 +222,15 @@ class InboundConfigMixin(object):
     Define the inbound-related required configuration.
     """
     inbound_branch = 'mozilla-inbound'
+    branch_path = 'integration'
```
wlach: Where is this variable actually used? I don't see anything functional in this patch actually referring to `branch_path`.

parkouss: Well, this is used just above, https://github.com/mozilla/mozregression/pull/319/files#diff-9fb3f9f996418622faaeaf680fa89a1dR84. This is the default value for https://github.com/mozilla/mozregression/pull/319/files#diff-4ac0cc6fa4a0d1bd5c2520a4969c04c4R232.

wlach: Ah, I see now. Makes sense.

PR #322, opened 2015-10-03
use an explicit message to use a number when --write-config ask for persist size

parkouss:


mozregression/config.py
```diff
@@ -78,8 +78,8 @@ def _get_persist_dir(default):
 def _get_persist_size_limit(default):
     print("You should choose a size limit for the persist dir. I recommend you"
           " to use %s GiB, so leave it blank to use that default. Else you"
-          " can set NONE to not limit the persist dir, or any custom value"
-          " you like." % default)
+          " can type NONE to not limit the persist dir, or any number you like"
+          " (a GiB value, so type 0.5 to allow ~500 Mo)." % default)
```
jay: Hi from bugzilla. I don't know what `Mo` is. Why not `MiB`?

parkouss: Perfectly fine for me, thanks! I fix it right now - Maybe I should have asked you a review ?

parkouss: I believe it is a french notation - thanks for telling me. :)

PR #324, opened 2015-10-04
Added instructions to install qt-creator on OS X

kapy2010:


parkouss:
Thanks for doing this! :+1:

PR #331, opened 2015-10-18
Offer a way to launch a specific release

MikeLing:


MikeLing:
Due to I recreate my PR from the other branch, I think we could close this PR anyway.

PR #332, opened 2015-10-19
 Bug 1199902 - use python logging instead of mozlog

parkouss:


parkouss:
We don't need that anymore, since we are going to keep mozlog.

PR #334, opened 2015-10-20
Bug 1214192 - Offer a way to launch a specific release

MikeLing:


mozregression/cli.py
```diff
@@ -483,7 +485,12 @@ def validate(self):
                 options.launch = parse_date(options.launch)
                 self.action = "launch_nightlies"
             except DateFormatError:
-                self.action = "launch_inbound"
+                try:
+                    options.release_date = date_of_release(int(options.launch))
```
parkouss: So we discussed about that, this will fail with ValueError if we use a real changeset.

mozregression/cli.py
```diff
@@ -258,7 +259,8 @@ def create_parser(defaults):
     parser.add_argument('--launch',
                         metavar="DATE_OR_REV",
                         help="Launch only one specific build by date (nightly)"
-                             " or changeset (inbound).")
+                             " or changeset (inbound). You can also launch a"
+                             " release by specific release number.")
```
parkouss: Nit: "You can also launch a release by specific release number" => "You can also launch a release using a specific release number"

parkouss:
Pretty good! Just a few things to fix:

- nit: https://travis-ci.org/mozilla/mozregression/builds/86753285#L283
- in the date_of_release function, there is a minor issue (I will comment)

mozregression/releases.py
```diff
@@ -58,8 +58,10 @@ def date_of_release(release):
     """
     Provide the date of a release.
     """
+    if isinstance(release, str):
+        release = ''.join([i for i in release if i.isdigit()])
```
parkouss: Hm, I don't like this test - if we have a revision, say "aeefd3ae4", it will detect that it is the release 34 :(

Just remove that code, and do an "except KeyError, ValueError" or write a specific try to test the int conversion.

mozregression/cli.py
```diff
@@ -483,7 +485,12 @@ def validate(self):
                 options.launch = parse_date(options.launch)
                 self.action = "launch_nightlies"
             except DateFormatError:
-                self.action = "launch_inbound"
+                try:
+                    options.release_date = date_of_release(options.launch)
+                    options.launch = parse_date(options.release_date)
```
parkouss: Can't we just write
```
options.launch = parse_date(date_of_release(options.launch))
```

mozregression/cli.py
```diff
@@ -258,7 +259,8 @@ def create_parser(defaults):
     parser.add_argument('--launch',
                         metavar="DATE_OR_REV",
                         help="Launch only one specific build by date (nightly)"
-                             " or changeset (inbound).")
+                             " or changeset (inbound). You can also launch a "
```
parkouss: nit: please put the final whitespace at the beginning of the next line.

parkouss:
Thanks for the tests!

mozregression/releases.py
```diff
@@ -59,8 +59,8 @@ def date_of_release(release):
     Provide the date of a release.
     """
     try:
-        return releases()[release]
-    except KeyError:
+        return releases()[int(release)]
+    except KeyError, ValueError:
```
parkouss: As stated in https://bugzilla.mozilla.org/show_bug.cgi?id=1214192, this is not doing what you want.

tests/unit/test_cli.py
```diff
@@ -193,6 +193,15 @@ def test_with_releases():
...
 
 
+@pytest.mark.parametrize('arg', [
+    '--launch=34', '--launch=reded3rt4'
+])
+def test_launch_with_release_number(arg):
+    config = do_cli(arg)
+    assert config.action == 'launch_nightlies'
+    assert config.options.launch == cli.parse_date(releases()[34])
+
```
parkouss: This test seems broken - if you pass '--launch=reded3rt4', config.options.launch should NOT be equal to cli.parse_date(releases()[34]).

I think you can remove the parametrize here. Just test for a right hardcoded value ('--launch=34'). the two asserts are fine though :)

I don't think you should test for a bad release number (like '--launch=reded3rt4') since it is equivalent to testing a good changeset (and that you tested it in another test). :) but if you want to do that, write the test in a new function test_launch_with_bad_release_number.

PR #340, opened 2015-10-26
Bug 1211147 - [gui] allow to use debug builds in the gui

MikeLing:


gui/mozregui/wizard.py
```diff
@@ -185,9 +186,21 @@ class InboundPage(WizardSelectionRangePage):
...
+        WizardPage.__init__(self)
+        self.build_type_model = QStringListModel()
+        self.ui.build_type.setModel(self.build_type_model)
+
+    def initializePage(self):
+        WizardSelectionRangePage.initializePage(self)
+        application_name = self.field('application').toPyObject()
+        self.build_type_model = QStringListModel(
+            [FC_REGISTRY.get(str(application_name)).BUILD_TYPES])
+
```
MikeLing: @parkouss  I'm curious about why it keeps telling me  no attribute 'BUILD_TYPES'

  File "/Users/mikeling/mozregression/gui/mozregui/wizard.py", line 202, in initializePage
    [FC_REGISTRY.get(str(application_name)).BUILD_TYPES])
AttributeError: type object 'FirefoxConfig' has no attribute 'BUILD_TYPES'

gui/mozregui/ui/inbound.ui
```diff
@@ -56,6 +56,20 @@
...
+      <widget class="QLabel" name="label_4">
+       <property name="text">
+        <string>Build Type</string>
+       </property>
+      </widget>
+     </item>
+     <item row="2" column="1">
+      <widget class="QComboBox" name="build_type">
+       <property name="toolTip">
+        <string>Choose a build type you want, notice only opt is available most of time</string>
```
parkouss: Please remove "notice only opt is available most of time". (we handle every case for inbound, the warning was for users using command lines and nightlies).

gui/mozregui/wizard.py
```diff
@@ -244,6 +256,8 @@ def options(self):
         else:
             kind = "changeset"
             fetch_config.set_inbound_branch(options['inbound_branch'])
+        if options['build_type'] is not 'opt':
```
parkouss: Hm, there is an issue there. If we only choose a nightly build, options['build_type'] will be '' and that will fail. Plus we don't need to make an exception for opt. Please just replace:
```
if options['build_type'] is not 'opt':
```
with:
```
if options['build_type']:
```

MikeLing:
Done! Thank you for your help @parkouss

parkouss:
Yup, thank you! Merged manually in https://github.com/mozilla/mozregression/commit/40b34f3629a21009435c67b95473a330929aaff6.

PR #343, opened 2015-10-30
Fix bug 1195296

ShivamNegi:
To create a profile in case of a non existing profile specified by the user, these two lines have been added in /mozregression/launchers.py. Which creates the directory corresponding to the profile path specified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ dist/*
 build/*
 .coverage
 htmlcov/
+mozregression.egg-info/
 
 gui/build/
 gui/dist/


mozregression/launchers.py
```diff
@@ -163,6 +163,8 @@ def _install(self, dest):
             raise
 
     def _start(self, profile=None, addons=(), cmdargs=(), preferences=None):
+        if not os.path.exists(profile):
```
parkouss: profile can be None - so it should be something like "if profile and not os.path.exists(profile)".

ShivamNegi: Made the suggested change.

parkouss:
Thanks @ShivamNegi, we're almost there!

Well first thing I would like to see the dir creation in create_profile, just after https://github.com/mozilla/mozregression/pull/343/files#diff-9be3d66461b771098aa1bcc305505480R129 for example - this way the fennec launcher will also benefits from the fix (sorry I missed that first time). And this make sense to have the create_profile do this task.

Second thing is a small nit - we run flake8 on travis-ci (see check panel above) and you have a small trailing whitespace: https://travis-ci.org/mozilla/mozregression/builds/88289171#L281

Can you fix that ? Thanks for working on this bug. :)

PR #347, opened 2015-11-12
Fix Unicode display and edit

xennygrimmato:
This is a fix for Bug 1224083 - https://bugzilla.mozilla.org/show_bug.cgi?id=1224083 .

On line 39, the new_value object was a QVariant object, which can be converted to a QString.
Using QString's __str__( ) method seems to have solved the issue.

### DIFF
--- a/gui/mozregui/pref_editor.py
+++ b/gui/mozregui/pref_editor.py
@@ -30,13 +30,16 @@ class PreferencesModel(QAbstractTableModel):
             if index.column() == 0:
                 return name
             else:
-                return repr(value)
+                if isinstance(value, basestring):
+                    return '"' + value + '"'
+                else:
+                    return value
 
     def flags(self, index):
         return Qt.ItemIsSelectable | Qt.ItemIsEditable | Qt.ItemIsEnabled
 
     def setData(self, index, new_value, role=Qt.EditRole):
-        new_value = str(new_value.toString())
+        new_value = unicode(new_value.toString())
         name, value = self.prefs[index.row()]
         if index.column() == 0:
             # change pref name


gui/mozregui/pref_editor.py
```diff
@@ -30,13 +30,13 @@ def data(self, index, role=Qt.DisplayRole):
             if index.column() == 0:
                 return name
             else:
-                return repr(value)
+                return QString(value).__str__()
```
parkouss: Well, I would still like to see the quotes, so repr(value) is fine for me. What do you think of removing the first 'u' (if there is one) in the result of repr(value) ?

gui/mozregui/pref_editor.py
```diff
@@ -30,13 +30,13 @@ def data(self, index, role=Qt.DisplayRole):
...
 
     def flags(self, index):
         return Qt.ItemIsSelectable | Qt.ItemIsEditable | Qt.ItemIsEnabled
 
     def setData(self, index, new_value, role=Qt.EditRole):
-        new_value = str(new_value.toString())
+        new_value = new_value.toPyObject().__str__()
```
parkouss: I believe this is the same, that is new_value will still be a 'str' object, not 'unicode', no ? I think just replacing str(new_value.toString()) to **unicode**(new_value.toString()) will work well. You can try it by editing a cell with a non unicode character (like "é").

xennygrimmato: new_value is a 'unicode' object now.
> "replacing str(new_value.toString()) to unicode(new_value.toString()) will work well."
Yes, that works fine too.

gui/mozregui/pref_editor.py
xennygrimmato: Using repr(value) doesn't render the non-ascii characters properly. They are rendered as '\xYZ' characters.

gui/mozregui/pref_editor.py
parkouss: Oh cool, if that is a unicode I'm not agains it. I find unicode() easier to read, but I'm ok with the other notation (maybe with a comment in this case to say that it is real unicode).

xennygrimmato: Maybe, unicode(new_value.toString()) would be better for readability and I must change it. What do you think? The current method hides this detail, that it is 'unicode'.

gui/mozregui/pref_editor.py
parkouss: eh, ok - thanks for pointing this out! Maybe something like ('"%s"' % value) would do it then if the value is of type string ? The reason why I would like to see quotes is because it can be a number, or a string representing a number, e.g "5" vs 5. But that's not that important, so we can leave out the quotes - as you prefer.

gui/mozregui/pref_editor.py
parkouss: I'm totally fine with unicode(new_value.toString()) - I find it better for readability too.

gui/mozregui/pref_editor.py
xennygrimmato: That's ok, I'll add the quotes.

xennygrimmato:
Made the changes as discussed above in commit 8e5c74b

xennygrimmato:
Latest changes: 5ac9571

gui/mozregui/pref_editor.py
```diff
@@ -30,13 +30,16 @@ def data(self, index, role=Qt.DisplayRole):
             if index.column() == 0:
                 return name
             else:
-                return repr(value)
+                if isinstance(value, basestring):
+                    return '"' + QString(value).__str__() + '"'
```
parkouss: can't we write just '"' + value + '"' since we know that it is a string now ?

xennygrimmato: Yes, we can. Made the changes and added in 70583ea

parkouss:
Small nit: https://travis-ci.org/mozilla/mozregression/builds/90773702#L334
QString is imported but not used now.

xennygrimmato:
Is it fine now? I will squash the commits.

parkouss:
Please do! A commit like "Bug 1224083 - fix preference display / unicode handling" would be fine (please start it with Bug 1224083). I'll be happy to merge this!

PR #350, opened 2015-11-17
Bug for persist folder fixed

GopianiS:
Creates a folder at initialisation of the of the DownloadManager class

### DIFF
--- a/mozregression/download_manager.py
+++ b/mozregression/download_manager.py
@@ -1,4 +1,4 @@
-import tempfile
+#import tempfile
 import threading
 import requests
 import os
@@ -229,7 +229,7 @@ class DownloadManager(object):
         self.persist_limit = persist_limit or PersistLimit(0)
         self.persist_limit.register_dir_content(self.destdir)
 
-        #if folder does not exist, create it
+        # if folder does not exist, create it
         if not os.path.isdir(destdir):
             os.makedirs(destdir)
 
--- a/mozregression/download_manager.py
+++ b/mozregression/download_manager.py
@@ -1,4 +1,4 @@
-#import tempfile
+import tempfile
 import threading
 import requests
 import os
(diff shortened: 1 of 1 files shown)


mozregression/download_manager.py
```diff
@@ -1,4 +1,4 @@
-import tempfile
+#import tempfile
```
parkouss: Hm, tempfile is still required I guess.

mozregression/download_manager.py
```diff
@@ -149,6 +149,7 @@ def _download(self, url, dest, finished_callback, chunk_size, session):
         # abruptly)
         temp = None
         bytes_so_far = 0
+
```
parkouss: please remove that blank line, it does not belong to this patch.

mozregression/download_manager.py
```diff
@@ -218,6 +219,7 @@ def download_things(manager):
     :param persist_limit: an instance of :class:`PersistLimit`, to allow
                           limiting the size of the download dir. Defaults
                           to None, meaning no limit.
+
```
parkouss: same here.

mozregression/download_manager.py
```diff
@@ -254,6 +260,7 @@ def download(self, url, fname):
         if a download is already running for the given fname, it is just
         returned. Else the download is created, started and returned.
         """
+
```
parkouss: same, remove the blank line please.

mozregression/download_manager.py
```diff
@@ -227,6 +229,10 @@ def __init__(self, destdir, session=requests, persist_limit=None):
         self.persist_limit = persist_limit or PersistLimit(0)
         self.persist_limit.register_dir_content(self.destdir)
 
+        # if folder does not exist, create it
+        if not os.path.isdir(destdir):
+            os.makedirs(destdir)
+
```
parkouss: looks all good to me here. Could you just adjust the comment to:
# if **persist**folder does not exist, create it
please ? that would allow to understand better why we are doing that, for when we will look at the code later.

parkouss:
Thanks @GopianiS ! Just squashed the commit and merged that in https://github.com/mozilla/mozregression/commit/34825b79ebe546eb282e92b22af3177e89dde695.

PR #352, opened 2015-11-18
Process logging

parkouss:


mozregression/cli.py
```diff
@@ -265,6 +265,12 @@ def create_parser(defaults):
                              " release using a specific release number, and"
                              " a nightly using a specific buildid")
 
+    parser.add_argument('-P', '--process-output', choices=('0', '1'),
+                        default=defaults['process-output'],
+                        help=("Manage process output logging. 1 to enable, 0"
```
wlach: I feel like `process output logging` is not entirely clear. I would at least say `Manage process output (stdout) logging.`

Also, why isn't this a store_true argument? I don't see why specifying 1 or 0 on the command line should really be necessary.

parkouss: Basically to allow the user to enable this option in the config file, but still allowing him to disable it from the command line.

store_true is not something that can be easily overridden from the command line when you have it defined in the config file. it would require a -\-no-output to deactivate it, and I don't see any simple way to manage it with argparse. I tend to use now argument that always take a parameter, so we can override the value easily when it is defined in the config file.
Make sense ? If you have another idea, tell me. :)

wlach: Could you make the option store_true or store_false depending on what the config file says?

parkouss: Yes. That would be weird to me, cmdline options depending on the config file... What if I just replace "1" and "0" possible values (yeah, that is quite horrible now that I see that again) to "stdout" and "none" ? There would be some space in case we want to add another new action, and "-P=stout" looks ok to me.
Will ?

parkouss:
Closed because the code is in #353.

PR #354, opened 2015-11-20
Fix back skip

parkouss:


mozregression/history.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    An history bisection step.
+    """
+    def __init__(self, build_range, index, verdict=None):
+        self.build_range = build_range
+        self.index = index
+        self.verdict = verdict
+
+
+class BisectionHistory(object):
```
wlach: You might consider making this class inherit from list, and just adding one or two custom methods for convenience. Something like:

```
class BisectionHistory(list):
    def add_step(self, build_range, index, verdict):
        self.append(BisectionStep(...))
```

mozregression/history.py
```diff
@@ -0,0 +1,46 @@
+"""
+Representation of the bisection history.
+"""
+
+
+class BisectionStep(object):
```
wlach: Could this be a namedtuple? https://docs.python.org/2.7/library/collections.html#namedtuple-factory-function-for-tuples-with-named-fields

parkouss: Sure, good idea.

mozregression/history.py
parkouss: Will do!

PR #358, opened 2015-11-27
Update documentation

parkouss:
Some minor updates for config file usage, and added two options in usage
section.


parkouss:
@wlach , could you review that please ?

PR #362, opened 2015-12-02
Gui inbound nightly

parkouss:


mozregression/fetch_configs.py
```diff
@@ -135,6 +136,25 @@ def set_build_type(self, build_type):
...
+        If not set or set to None, default repos would be used (see
+        :meth:`get_nightly_repo` and :attr:`inbound_branch`)
+        """
+        self.repo = branches.get_name(repo) if repo else None
+
+    def should_use_taskcluster(self):
+        """
+        Returns true if taskcluster should be used as the bisection method.
+
+        Should be only called when the repo and the build type are set.
```
wlach: Do you want to add an assertion about this?

wlach: (might help for debugging)

mozregression/fetch_configs.py
```diff
@@ -135,6 +136,25 @@ def set_build_type(self, build_type):
...
+        Allow to define the repo name.
+
+        If not set or set to None, default repos would be used (see
+        :meth:`get_nightly_repo` and :attr:`inbound_branch`)
+        """
+        self.repo = branches.get_name(repo) if repo else None
+
+    def should_use_taskcluster(self):
+        """
+        Returns true if taskcluster should be used as the bisection method.
```
wlach: true -> True

mozregression/fetch_configs.py
parkouss: Well, in fact it is totally valid to have those not set (e.g. None or 'opt') - this only means that no specific branch or build type has been defined by the user.
I wanted to say that this method relies on those values to return a correct result (if you call it, then change the repo and call it again, the return value may change).
I'll change the comment - thanks for the pointer!

mozregression/fetch_configs.py
parkouss: Will do, thanks :)

PR #364, opened 2015-12-04
clean automation run scripts and add some setup documentation

parkouss:


check.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python
+"""
+Run flake8 checks and tests.
+"""
```
wlach: I would personally write this as a shell script, but it's up to you.

parkouss: Nah - unfortunately that run on windows also;

PR #366, opened 2015-12-06
Gui single build

parkouss:


gui/mozregui/build_runner.py
```diff
@@ -0,0 +1,187 @@
...
+            pass
+        BuildDownloadManager._download_finished(self, task)
+
+    def focus_download(self, build_info):
+        build_url, fname = self._extract_download_info(build_info)
+        dest = self.get_dest(fname)
+        build_info.build_file = dest
+        # first, stop all downloads in background (except the one for this
+        # build if any)
+        self.cancel(cancel_if=lambda dl: dest != dl.get_dest())
```
MikeLing: WOW! This part of code is awesome, I haven't use lambda before even I had know it for a while!

gui/mozregui/report.py
```diff
@@ -122,6 +123,7 @@ def attach_bisector(self, bisector):
                                self)
 
         if bisector:
+            self.attach_single_runner(None)
             _bulk_action_slots('connect',
```
MikeLing: Why we make it as non-public variables, do we have any necessary to do so?

PR #367, opened 2015-12-11
Bug 1231745 - fix race condition when removing dl dir

parkouss:
Ensure that we don't have a race condition when the download_dir is
removed by first waiting for download threads to be finished.

### DIFF
--- a/mozregression/download_manager.py
+++ b/mozregression/download_manager.py
@@ -250,6 +250,13 @@ class DownloadManager(object):
                     if download.is_running():
                         download.cancel()
 
+    def wait(self, raise_if_error=True):
+        """
+        Wait for all downloads to be finished.
+        """
+        for download in self._downloads.values():
+            download.wait(raise_if_error=raise_if_error)
+
     def download(self, url, fname):
         """
         Returns a started download instance, or None if fname is already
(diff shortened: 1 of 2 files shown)


mozregression/download_manager.py
```diff
@@ -234,7 +234,7 @@ def __init__(self, destdir, session=requests, persist_limit=None):
     def get_dest(self, fname):
         return os.path.join(self.destdir, fname)
 
-    def cancel(self, cancel_if=None):
+    def cancel(self, cancel_if=None, wait=False):
```
wlach: I'm not sure if I understand why this is optional? When would we not want wait to be `True`?

parkouss: Because this may block (in case the network is busy or something), and it is used in the gui, https://github.com/mozilla/mozregression/blob/91dc13a55125eb72db810f3eca4ededcc6b8f5ab/gui/mozregui/build_runner.py#L166

I don't want the GUI to be blocking. It works in the gui because we wait (after the mainwindow is closed) for thread completion, then we remove the folder if required, https://github.com/mozilla/mozregression/blob/master/gui/mozregui/main.py#L148-L150. this is synchronized and blocking - but done when no window is shown so the user usually don't notice it.

Maybe I can let this function as it was, and just use the new wait() call in the commandline main.py file - this can be a bit more readable.

@wlach thoughts or suggestions ?

wlach: Yeah I think that might be a better idea. Put the wait in the commandline file with a comment as to why it's there (maybe even linking to this bug #)

PR #370, opened 2015-12-14
implemented auto-complete for repository field

GopianiS:


parkouss:
@GopianiS, thanks for the work here! I reused part of it in https://github.com/mozilla/mozregression/pull/412. I'm closing this since it will be handled from there.

PR #373, opened 2015-12-16
[GUI] Added support for profile-persistence

crisron:
Added a combobox in the Profile Page from which the `profile-persistence` can be selected. The default is set as `clone`.


gui/mozregui/build_runner.py
```diff
@@ -74,7 +74,9 @@ def evaluate(self, build_info, allow_back=False):
         else:
             self.evaluate_started.emit('')
 
-    def finish(self, verdict):
+    def finish(self, verdict, options):
+        if options['profile'] and options['profile_persistence'] == 'clone-first':
+            options['profile'].cleanup()
```
parkouss: I think this should be not be called in the launcher.finish method, because this is called after every tested build. I believe you have to put that code in Bisector.stop instead. Also, does "if options['profile']" works in case the profile field is empty ? (Just a question)

gui/mozregui/wizard.py
```diff
@@ -228,6 +244,17 @@ def options(self):
         fetch_config.set_repo(options['repository'])
         fetch_config.set_build_type(options['build_type'])
 
+        # create a profile if required
+        launcher_class = LAUNCHER_REGISTRY.get(fetch_config.app_name)
+        launcher_class.check_is_runnable()
```
parkouss: check_is_runnable should not be called here - it does some check to know if a launcher is runnable (for example for fennec - firefoxon android -, ensure that a phone is connected. Just remove that line, this is checked elsewhere.

parkouss:
@crisron thanks for the work!

There is some flake8 nits that needs to be fixed - you can see that the travis-ci and appveyor builds above are red (failing), here is a quick link to the errors:
https://travis-ci.org/mozilla/mozregression/builds/97294368#L634-L657
Can you fix that please ?

Also I wrote some comments, a few things that should be addressed/answered - then we're good to land I think!

gui/mozregui/build_runner.py
crisron: Yes, I would change that. Yes, it will work. It will evaluate to `None` in that case.

parkouss:
Merged in https://github.com/mozilla/mozregression/commit/252197a78ee8f77fb80cd74a21a3f5fc08ac0efb. Thanks @crisron !

PR #375, opened 2015-12-18
[gui] Bisection Informations context menu text turns white when hovered

crisron:
Removed the unnecessary setStyleSheet() call


gui/mozregui/report.py
```diff
@@ -299,7 +299,7 @@ def update_content(self, item):
             return
 
         html = ""
-        self.setStyleSheet("background-color:%s;" % GRAY_WHITE.name())
+        print "Setting StyleSheet"
```
parkouss: print nit! :)

PR #376, opened 2015-12-20
Bug 1234029 - use debug builds from archives.m.o

parkouss:
For firefox and jsshell, provides access to debug builds using
archives.m.o.

Those builds are stored using 'debug' in their filename, and
using TaskCluster is not required anymore.


wlach:
Rebasing this in #465

PR #377, opened 2015-12-21
Use gecko v2 taskcluster routes

parkouss:


parkouss:
Closing, #406 will do the job.

PR #378, opened 2015-12-22
Tc authenticate cert

parkouss:


mozregression/tc_authenticate.py
```diff
@@ -0,0 +1,45 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
```
wlach: Not really a big deal, but I found out a while ago that we don't need to put this header in our source files.

See: https://bugzilla.mozilla.org/show_bug.cgi?id=1192801

PR #379, opened 2015-12-22
Bug 1227780 - addition of certificate

nhirata:
Initial PR, haven't tested yet.  
Need to test first...


nhirata:
Closing PR.  Julien has a better patch than I do which has a try for the certificate.

PR #381, opened 2015-12-22
[GUI] download background builds

parkouss:
Add the ability to download background builds (now the default)
and a global pref to change that.

### DIFF
--- a/gui/mozregui/ui/global_prefs.ui
+++ b/gui/mozregui/ui/global_prefs.ui
@@ -33,24 +33,24 @@
       <widget class="DirectorySelectWidget" name="persist" native="true"/>
      </item>
      <item row="1" column="0">
-         <widget class="QLabel" name="label">
-             <property name="toolTip">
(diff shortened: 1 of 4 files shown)


gui/mozregui/ui/global_prefs.ui
```diff
@@ -69,6 +69,23 @@
...
+       </property>
+      </widget>
+     </item>
+     <item row="3" column="1">
+      <widget class="QCheckBox" name="bg_downloads">
+       <property name="text">
+        <string/>
+       </property>
+      </widget>
+     </item>
```
wlach: Does this really need to be exposed in the UI?

parkouss: hm, maybe not, you're right. But since we have the option in the CLI, I guess it make sense to have it too.

Maybe we could do an 'advanced options' panel that can hide this option as a default ? And clicking on 'advanced options' would show it. I think other options can be in there, like the http timeout.

PR #383, opened 2015-12-22
Fix resume info to specify 'good' and 'bad' rather than 'good-rev' an…

staktrace:
…d 'bad-rev' which have been removed.


parkouss:
Thanks. :)

PR #384, opened 2015-12-24
Bug1234603 - [gui] Added Show/Hide Advanced Options button in global prefs dialog

crisron:
Added a 'Show Advanced Options' button in the global preferences dialog. Initially, `http timeout` and `background downloads` options are hidden. On clicking the button, the button's text changes to 'Hide Advanced Options' and the advanced options become visible.

### DIFF
--- a/gui/mozregui/global_prefs.py
+++ b/gui/mozregui/global_prefs.py
@@ -54,6 +54,23 @@ class ChangePrefsDialog(QDialog):
         self.ui.http_timeout.setValue(options['http_timeout'])
         self.ui.persist_size_limit.setValue(options['persist_size_limit'])
         self.ui.bg_downloads.setChecked(options['background_downloads'])
+        self.ui.advanced_options.setText("Show Advanced Options")
+        self.toggle_visibility(False)
+        self.ui.advanced_options.clicked.connect(self.toggle_adv_options)
+
+    def toggle_adv_options(self):
+        if self.ui.advanced_options.text() == "Show Advanced Options":
+            self.ui.advanced_options.setText("Hide Advanced Options")
+            self.toggle_visibility(True)
+        else:
+            self.ui.advanced_options.setText("Show Advanced Options")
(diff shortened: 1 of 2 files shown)


gui/mozregui/global_prefs.py
```diff
@@ -54,6 +55,29 @@ def __init__(self, parent=None):
...
+
+    def toggle_adv_options(self):
+        if self.ui.advanced_options.text() == QString("Show Advanced Options"):
+            self.ui.advanced_options.setText(QString("Hide Advanced Options"))
+            self.show_adv_options()
+        else:
+            self.ui.advanced_options.setText(QString("Show Advanced Options"))
+            self.hide_adv_options()
+
+    def show_adv_options(self):
```
parkouss: Could you rewrite this using setVisible(bool) instead of show()/hide() ? You will only need one function then to hide or show the items instead of two.

parkouss:
So, you should still use the "form layout" that was here before. Try to resize the window without your patch - and try with. "form layout" allow to automatically handle the resizing, and I believe we need it here.

I think for th UI, you should only add a the new button outside the layout - and having this button at the bottom (live the two others, but on the left side) would be nice.

gui/mozregui/global_prefs.py
```diff
@@ -54,6 +55,23 @@ def __init__(self, parent=None):
         self.ui.http_timeout.setValue(options['http_timeout'])
         self.ui.persist_size_limit.setValue(options['persist_size_limit'])
         self.ui.bg_downloads.setChecked(options['background_downloads'])
+        self.ui.advanced_options.setText(QString("Show Advanced Options"))
```
parkouss: You can use python strings, no need to call QString(). Could you change that ? I guess you won't need the QString import after that. (This is mainly to be consistent with the rest of the code, and to not add unnecessary code).
Sorry I missed that the first time :)

PR #388, opened 2015-12-31
Bug 1203694 - [gui] provide an installer or binary for mac osx

MikeLing:


parkouss:
Hey, look like you are making some serious progress here! Excellent @MikeLing !

The deployed file (https://github.com/MikeLing/mozregression/blob/bda9fde4a82ae278bab63cdd9b8609fa3334e7f9/.travis.yml#L52) needs to be different also (should be the new .dmg file). I *think* you need to create two deploy sections (https://docs.travis-ci.com/user/deployment/#Deploying-to-Multiple-Providers) and for each add a condition on the os (https://docs.travis-ci.com/user/deployment/#Conditional-Releases-with-on%3A, https://github.com/numenta/nupic/blob/master/.travis.yml#L86)

You can try that everything works by creating and pushing tags on your fork to trigger the builds. Basically:
```
git tag -a gui-test-1
git push --tags
```

After the build is done, on your repo you should see new releases, and the files should have been uploaded, and should work! You probaly will need to change the value of the apikey temporarily (https://github.com/MikeLing/mozregression/blob/bda9fde4a82ae278bab63cdd9b8609fa3334e7f9/.travis.yml#L51) so that it can work on your repo. I don't remember exactly how I did that, I believe it is in github settings personal access token.

I'm excited to see this fully working!

MikeLing:
Hi @parkouss, I'm not sure why this happen https://travis-ci.org/mozilla/mozregression/jobs/99780684#L324. I found same error on osx but with different error https://travis-ci.org/mozilla/mozregression/jobs/99780685#L3040. And that happened on my local environment also. Could you give me some hit about it? And I will try to figure it out. Thank you :)

MikeLing:
I think these error message may be helpful https://pastebin.mozilla.org/8855719

parkouss:
There is at least one thing that should be adressed, the coverage should only be applied to Linux. So for https://github.com/MikeLing/mozregression/blob/78cd766c00c8d0bee57aa69e4e505ec3de25d946/.travis.yml#L37, you should remove the --with-coverage flag (for osx only) and also not do https://github.com/MikeLing/mozregression/blob/78cd766c00c8d0bee57aa69e4e505ec3de25d946/.travis.yml#L40-L41 for osx. Tell me if that helps :)

MikeLing:
Hmm, It still goes wrong with same error.  And it still requires PyQt4(which is I had install already). :(

MikeLing:
And I haven't found .dmg on my release branch, even the build on osx is green.
I found these messages on the osx build session:
The command "if [ "$OS" = "Linux" ]; then sudo ./check.py --with-coverage --with-gui; fi" exited with 0.
$ if ["$OS" = "MacOSX" ]; then sudo ./check.py --with-gui; fi
**/Users/travis/build.sh: line 45: [MacOSX: command not found**
It confuse me a lot! where this "command not found comes from"?

.travis.yml
```diff
@@ -1,37 +1,45 @@
...
 python:
-    - "2.7"
+- '2.7'
 virtualenv:
-    # required because we install the python-qt4 ubuntu package
-    system_site_packages: true
-
+  system_site_packages: true
+before_install:
+- OS=$(if [[ "$TRAVIS_OS_NAME" = "osx" ]]; then echo "MacOSX"; else echo "Linux";
```
parkouss: I think this OS variable is causing the error, remove that line and try using TRAVIS_OS_NAME only in this file.

gui/build.py
```diff
@@ -134,6 +135,11 @@ def do_bundle(options):
     if IS_WIN:
         makensis_path = os.path.join(options.nsis_path, "makensis.exe")
         call(makensis_path, 'wininst.nsi', cwd='wininst')
+    elif IS_MAC:
+        arg = ['hdiutil', 'create', 'dist/mozregression-gui.dmg', '-srcfolder']
+        arg.append('dist/')
+        arg.append('-ov')
+        call(*arg)
```
parkouss: Nit: Since you are not dynamically changing args, you can just do something like:
```python
call('hdiutil', 'create', 'dist/mozregression-gui.dmg', '-srcfolder', 'dist/', '-ov')
```

.travis.yml
```diff
@@ -1,37 +1,64 @@
...
 
-python:
-    - "2.7"
-virtualenv:
-    # required because we install the python-qt4 ubuntu package
-    system_site_packages: true
+before_install:
+    - if [ "$TRAVIS_OS_NAME" == "osx" ]; then
+         brew update; brew outdated xctool || brew upgrade xctool;
+         pip install virtualenv;
```
parkouss: Could you move thos commands in the install section, just before the other brew calls ?
I think it will be more readable to have every brew call at the same place, and we could remove the before_install section then.

parkouss:
Great! Merged manually in https://github.com/mozilla/mozregression/commit/edb8834e0f02996f595b967ca9d7fc8faad23fc5, to change back the API keys.

PR #390, opened 2016-01-05
Bug 1199902 - use mozlog proxy loggers

parkouss:


mozregression/bisector.py
```diff
@@ -14,6 +14,8 @@
 from mozregression.branches import find_branch_in_merge_commit
 from mozregression.json_pushes import JsonPushes
 
+LOG = get_proxy_logger('Bisector')
```
wlach: I'd probably use `logger` instead of `LOG` (that seems to be the default in most python code I've seen), but it's up to you. :)

parkouss: oups, missed that. :(
Oh well, I think it's ok like this.

PR #395, opened 2016-01-11
ADBHost().devices() now raises ADBError instead of OSError

saru95:
According to https://bugzilla.mozilla.org/show_bug.cgi?id=1236469#c3

### DIFF
--- a/mozregression/launchers.py
+++ b/mozregression/launchers.py
@@ -291,9 +291,8 @@ class FennecLauncher(Launcher):
     def check_is_runnable(cls):
         try:
             devices = ADBHost().devices()
-        except ADBError:
-            raise LauncherNotRunnable("mozdevice.adb.ADBError: [Errno 2] No"
-                                      " such file or directory: adb is not executable.")
+        except ADBError as adb_error:
+            raise LauncherNotRunnable(str(adb_error))
         if not devices:
             raise LauncherNotRunnable("No android device connected."
                                       " Connect a device and try again.")
--- a/mozregression/launchers.py
+++ b/mozregression/launchers.py
@@ -289,13 +289,13 @@ class FennecLauncher(Launcher):
 
     @classmethod
     def check_is_runnable(cls):
-        # ADBHost().devices() seems to raise OSError when adb is not
-        # installed and in PATH. TODO: maybe fix this in mozdevice.
+        # ADBHost().devices() seems to raise ADBError
+        # when adb is not executable or not in PATH.
         try:
             devices = ADBHost().devices()
-        except OSError:
+        except ADBError:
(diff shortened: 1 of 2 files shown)


mozregression/launchers.py
```diff
@@ -289,13 +289,13 @@ class FennecLauncher(Launcher):
 
     @classmethod
     def check_is_runnable(cls):
-        # ADBHost().devices() seems to raise OSError when adb is not
-        # installed and in PATH. TODO: maybe fix this in mozdevice.
+        # ADBHost().devices() seems to raise ADBError
+        # when adb is not executable or not in PATH.
```
parkouss: I would be fine to just remove that comment now - it is part of the contract that ADBHost().devices() may raise ADBError now, so the comment is not that helpful anymore.

mozregression/launchers.py
```diff
@@ -289,13 +289,13 @@ class FennecLauncher(Launcher):
...
         try:
             devices = ADBHost().devices()
-        except OSError:
+        except ADBError:
             raise LauncherNotRunnable("adb (Android Debug Bridge) is not"
-                                      " installed or not in the PATH.")
+                                      " executable or not in the PATH.")
```
parkouss: So the idea is to raise LauncherNotRunnable - but with the message contained inside ADBError exception. Because ADBError may contain (now or in the future) another reason for the failure. Can you change that ?

mozregression/launchers.py
saru95: Should I submit a new PR ? Sorry for the trouble , I'm a beginner .

mozregression/launchers.py
saru95: Yeah , I'll submit a new PR . would this do , "mozdevice.adb.ADBError: [Errno 2] No such file or directory: adb is not executable." ?

mozregression/launchers.py
parkouss: No problem! No, you should just add a new commit with the fixes. I will later squash the commit.

mozregression/launchers.py
parkouss: So, first, avoid creating a new PR - just add new commits and push them in this PR.
Then, you should not write a string to explain the failure, because we want the string contained inside the ADBError exception that we intercept.

In other words, the idea is to create a new error that should contain the same text as the first error.

parkouss: Oh, if you want, we can chat on irc, I am parkouss in the \#ateam channel. You can connect with mibbit for example:
https://wiki.mozilla.org/IRC#Mibbit

parkouss:
Landed and rebased in https://github.com/mozilla/mozregression/commit/2e3037f0d7a384877f2dd0c42c4da7410b6e2d22. Thanks @saru95

PR #397, opened 2016-01-12
Remove unused shebangs

JohanLorenzo:
We already have the [shebang at the top of the file](https://github.com/JohanLorenzo/mozregression/blob/adc1f344959abf1bd5c5847922f3ba478220e1ce/mozregression/main.py#L1). Then I guess we just need to change the permissions :smiley:

### DIFF
--- a/mozregression/bisector.py
+++ b/mozregression/bisector.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
-
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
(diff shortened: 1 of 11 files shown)


parkouss:
Oh, the shebang is there, but never used (we could remove it btw). We always install mozregression using pip, so it install a mozregression executable. (if you use virtualenv, it ends up in venv/bin).

Why do you want it to be executable ?

JohanLorenzo:
Discussed on IRC. I used executable files in order to make sure I was running the latest version of the code (without having to update the virtualenv each time). @parkouss said you can just use 
``` sh
pip install -e .
```

Then `mozregression` at the virtualenv level will always be at latest changes.

Let's use this PR to remove the unused shebangs.

JohanLorenzo:
I verified the files were supposed to be not executable. Now, there is one shebang left in `check.py`

parkouss:
Yeah, the file check.py is not distributed during installation step, so this is the only one that should keep the shebang.

parkouss:
Thanks!
Landed in https://github.com/mozilla/mozregression/commit/36a44256b35ebb3ef7329fcf5a42ac4f07384ba8.

PR #399, opened 2016-01-15
Bugfix 1173984 filter log levels

wasifhyder:
Implemented feature from bug-1173984.

The feature allows filtering of log levels by accessing the context menu and selecting the level. The different log levels are checkable, with the default being "debug".


parkouss:
Oh the build failures are unrelated to your PR. I just committed a fix for that, https://github.com/mozilla/mozregression/commit/d942599a37e8c92da83d58013f5c39c7f953a747. Could you rebase your PR ?

parkouss:
This is quite good! So basically one thing is missing: new messages that are under the current level should be invisible. For that, you need to keep the current level value on the instance (starting with the value for debug). In the slot to change the log level, you should also update that value, and in the slot where messages are appended you should hide them if needed (I believe you can write a method to hide/show a block given its data so you don't copy paste code).
I'll add a few comments in the PR. I really like the changes, thanks!

gui/mozregui/log_report.py
```diff
@@ -1,7 +1,9 @@
 from PyQt4.QtCore import QObject, pyqtSlot as Slot, pyqtSignal as Signal
 from PyQt4.QtGui import QPlainTextEdit, QTextCursor, QColor, \
-    QTextCharFormat
+    QTextCharFormat, QMenu, QAction, QTextBlock, QTextBlockUserData, QActionGroup
```
parkouss: This line is too long - there are a few other style nits in the patches, please run "./check.py -G" to see them.

gui/mozregui/log_report.py
```diff
@@ -27,15 +46,44 @@ def on_log_received(self, data):
         cursor_to_add = QTextCursor(message_document)
         cursor_to_add.movePosition(cursor_to_add.End)
         cursor_to_add.insertText(log_message + '\n')
+
         if data['level'] in COLORS:
             fmt = QTextCharFormat()
             fmt.setForeground(COLORS[data['level']])
             cursor_to_add.movePosition(cursor_to_add.PreviousBlock)
+            log_lvl_data = LogLevelData(log_levels[data['level'].upper()])
```
parkouss: I wonder if you should store directly the numeric value in there, it would be more efficient (you won't have to do the conversion each time a level change) ?

gui/mozregui/log_report.py
```diff
@@ -27,15 +46,44 @@ def on_log_received(self, data):
...
                                                       cursor_to_add.position())
             cursor_to_add_fmt.mergeCharFormat(fmt)
         self.ensureCursorVisible()
 
+    @Slot()
+    def on_custom_context_menu_requested(self):
+        menu = QMenu(self)
+        for action in self.actions:
+            menu.addAction(action)
+        #self.contextMenu.move(self.cursor().pos())
```
parkouss: If you don't use that, just remove the line.

gui/mozregui/log_report.py
parkouss: Oh, forget it, you ARE doing it already;

gui/mozregui/log_report.py
```diff
@@ -27,15 +46,44 @@ def on_log_received(self, data):
...
+        menu.popup(self.cursor().pos())
+
+    @Slot()
+    def on_log_filter(self):
+        log_lvl_name = str(self.sender().iconText()).upper()
+        log_lvl = log_levels[log_lvl_name]
+        cursor = QTextCursor(self.document())
+        current_block = cursor.block()
+        while True:
```
parkouss: Hm, I don't like this while True with the break, since it is based on the presence or not of the user data. You should be able to iterate over all the block with a syntax like:
```python
while current_block.isValid():
    # do stuff
    current_block = current_block.next()
```
This way you are sure that you get every block.

gui/mozregui/log_report.py
```diff
@@ -27,15 +46,44 @@ def on_log_received(self, data):
...
+            if current_block.userData():
+                block_log_lvl = current_block.userData().log_lvl
+                if block_log_lvl <= log_lvl:
+                    current_block.setVisible(True)
+                else:
+                    current_block.setVisible(False)
+                current_block = current_block.next()
+            else:
+                break
+        self.viewport().update()
```
parkouss: Just a question, I suppose it is required ?

gui/mozregui/ui/mainwindow.ui
```diff
@@ -158,6 +158,9 @@
     <layout class="QVBoxLayout" name="verticalLayout_3">
      <item>
       <widget class="LogView" name="log_view">
+       <property name="contextMenuPolicy">
+        <enum>Qt::CustomContextMenu</enum>
+       </property>
```
parkouss: This is cool to see this in the .ui. Thanks!

gui/mozregui/log_report.py
wasifhyder: What is required?

gui/mozregui/ui/mainwindow.ui
wasifhyder: (Y)

gui/mozregui/log_report.py
parkouss: The call
```
self.viewport().update()
```
(sorry, that was not really explicit!)

wasifhyder:
Hi, I just saw this.

I'll fix it as soon as possible.

gui/mozregui/log_report.py
wasifhyder: Ah, that's fine. 

The call updates the log view after changing the visibility on the blocks. Without doing so, the log view doesn't update unless a viewport update is triggered by some other part of the application.

parkouss: Thanks a lot for checking it. :)

gui/README.rst
```diff
@@ -67,7 +67,8 @@ a virtualenv here, just be sure to pass the *-\-system-site-packages* flag
 when you create it to be able to use the system-wide pyqt4. See this link
 (http://docs.python-guide.org/en/latest/dev/virtualenvs/) to learn more
 about python virtualenvs. You should also consider using virtualenvwrapper
-(https://virtualenvwrapper.readthedocs.org/en/latest/).
+(https://virtualenvwrapper.readthedocs.org/en/latest/).ls
```
parkouss: typo ? seems to me that it is a mistake, please remove this change (or otherwise tell me why it is here :))

gui/mozregui/log_report.py
```diff
@@ -1,16 +1,25 @@
 from PyQt4.QtCore import QObject, pyqtSlot as Slot, pyqtSignal as Signal
-from PyQt4.QtGui import QPlainTextEdit, QTextCursor, QColor, \
-    QTextCharFormat
+from PyQt4.QtGui import QPlainTextEdit, QTextCursor, QColor
+from PyQt4.QtGui import QTextCharFormat, QMenu, QAction
+from PyQt4.QtGui import QTextBlockUserData, QActionGroup
```
parkouss: I would prefer to use only one 'from PyQt4.QtGui' - if you don't like the backslashes (I understand;), you can use parenthesis, like:
```python
from PyQt4.QtGui import (QPlainTextEdit, QTextCursor, QColor
                                           QTextCharFormat, QMenu, QAction, ...)
```

parkouss: sorry for the bad indentation, this is not intended. :)

gui/tests/test_log_report.py
```diff
@@ -38,3 +38,37 @@ def test_log_report_report_no_more_than_1000_lines(log_view):
...
+    for current_block in log_view.text_blocks():
+        for log_level in log_report.log_levels.keys():
+            if log_level in current_block.text():
+                assert current_block.userData().log_lvl == \
+                       log_report.log_levels[log_level]
+
+
+def test_log_report_filters_data_below_current_log_level(log_view):
+    # Hardcoded for 1 log level
+    log_view.log_lvl = 2
```
parkouss: Hm, if you don't mind, I would prefer to see log_level['DEBUG'] or something instead of the hardcoded 2. Just in case mozlog changes its values, the code here won't break (and it will be more readable).

gui/tests/test_log_report.py
```diff
@@ -38,3 +38,37 @@ def test_log_report_report_no_more_than_1000_lines(log_view):
...
+    current_log_level = log_view.log_lvl
+    # Inserts a log message for each log user level
+    for log_level in log_report.log_levels.keys():
+        log_view.log_model({'message': "%s message" % log_level,
+                            'level': '%s' % log_level, 'time': time.time()})
+    # Check that log messages above the current log level are visible
+    # and log messages below the log level are invisible
+    print()
```
parkouss: nit: remove that print.

gui/README.rst
wasifhyder: Yeah, definitely typo. 

Fixed

gui/mozregui/log_report.py
wasifhyder: Done.

parkouss:
Merged in https://github.com/mozilla/mozregression/commit/002dc8221a19998af429337e3242ecb103826dec. Thanks!

PR #408, opened 2016-01-22
Bug 1241830 - [GUI] stopping a bisection can be quite long

parkouss:


mozregression/build_range.py
```diff
@@ -129,8 +129,16 @@ def mid_point(self):
         accessed. Note that this method may resize the build range if some
         builds are invalids (we were not able to load build_info for some
         index)
+
+        if `interrupt` is given, it should be a callable that takes no
+        arguments and returns True when you want to stop looking for the
+        mid_point. In this case, StopIteration will be raised. This is provided
+        because this methods may take a long time to finish, and callers may
+        want to end it at some point.
```
wlach: I wonder if it wouldn't make more sense to use a threading event object here?

https://docs.python.org/2/library/threading.html#event-objects

parkouss: Oh, I believe it is better/safer - though I won't put that in mozregression core (since there is no threads involved here), but in the callback method that is passed from the GUI I will change it to return the event.is_set() result.

PR #411, opened 2016-01-30
Bug 1241803 - use abc module where we can

MikeLing:


tests/unit/test_bisector.py
```diff
@@ -12,7 +12,13 @@
 from mozregression.errors import LauncherError, MozRegressionError
 
 
+def mock_abstractmethod():
+    raise NotImplementedError
+
```
MikeLing: Hi @parkouss , so I found a way to allow us to test the ABC which is patch the abstractmethod when we first init it. But I need to avoid call it before assign it to __abstractmethod__, because it will throw an exception anyway. like:
> tests/unit/test_test_runner.py:162: in <module>
>    class TestCommandTestRunner(unittest.TestCase):
> tests/unit/test_test_runner.py:164: in TestCommandTestRunner
>    @patch.multiple(test_runner.TestRunner, __abstractmethod__=mock_abstractmethod())
> tests/unit/test_test_runner.py:14: in mock_abstractmethod
>    raise NotImplementedError
> E   NotImplementedError
 
Do you have any ideal about it?

tests/unit/test_bisector.py
```diff
@@ -12,7 +12,13 @@
...
+
 class TestBisectorHandler(unittest.TestCase):
+
+    @patch.multiple(BisectorHandler, __abstractmethod__=mock_abstractmethod())
```
parkouss: Instead of all this magic, why don't create a **TestBisectorHandler** new class, that inherit form **BisectorHandler** and implement the required methods ? Then the tests can use **TestBisectorHandler** instead of **BisectorHandler**.

MikeLing: oh! You're totally right! Thank you for you advice :)

MikeLing:
Hi @parkouss , I got mock error like: https://pastebin.mozilla.org/8858309. Why  it keeps telling me 'mozregression.test_runner.ManualTestRunner' does not have the attribute 'create_launcher'?

MikeLing:
The pastebin link I given may outdate, here is the new one https://pastebin.mozilla.org/8858593 :)

MikeLing:
hmm, even we got some errors on travis and appveyor. I guess it's not about the PR itself. So I would like to ask review for now :)

mozregression/fetch_configs.py
```diff
@@ -194,7 +194,7 @@ class NightlyConfigMixin(object):
     Note that subclasses must implement :meth:`_get_nightly_repo` to
     provide a default value.
     """
-    archive_base_url = ARCHIVE_BASE_URL
```
parkouss: Hmm, bad merge ? :) We still need that line I think :)

mozregression/test_runner.py
```diff
@@ -13,37 +13,41 @@
...
+                    % build_info.build_date.strftime("%Y%m%d%H%M%S"))
+        else:
+            desc = "for %s" % build_info.build_date
+    else:
+        desc = ("built on %s, revision %s"
+                % (build_info.build_date,
+                   build_info.short_changeset))
+    LOG.info("Running %s build %s" % (build_info.repo_name, desc))
+
+    return mozlauncher(build_info)
```
parkouss: I was thinking, maybe we could implement that in launcher.create_launcher instead ? This way we would have only one function create_launcher. If this is too complex (I guess some tests will complain again :() then we can do that in a follow up.

parkouss:
This looks good overall, thanks for doing this. The error comes from a nit, see the first comment.

mozregression/test_runner.py
MikeLing: hmm, I had thought about this too. But it means we may need to import whole launcher like "from mozregression import launcher", however we only use creat_launcher for it.

We can merge this for now and to see if we need do some follow up work for it anyway :)

PR #416, opened 2016-02-04
Bug 1235003

wasifhyder:
Breaks the drop down into good & bad buttons as well as the drop down.


parkouss:
Thanks, merged in https://github.com/mozilla/mozregression/commit/69f1be7dec004493c2f6b1bafffc556de8b09d8a.

I did a few fixes (the travis and appveyor builds were red because of empty lines) and small modifications, see in the bug.

PR #422, opened 2016-02-22
drop stray space before question mark

dholbert:
If you use mozregression with a "good" date that's newer than the "bad" date, it asks you:
> Maybe you wanted to use the --find-fix flag ?

Note the awkward space character at the end of the sentence there, before the question mark.  This pull request just drops that stray space.


wlach:
Yes, this looks slightly better. Thanks!

PR #424, opened 2016-03-02
Update Readme to include getting started links.

mhoye:


parkouss:
Thanks @mhoye, looks good.

PR #426, opened 2016-03-09
Bug 1254804 - mozregression doesn't properly visit inbound builds

parkouss:
Try to automatically bisect on inbounds when we don't have enough
information to know where a merge comes from.

### DIFF
--- a/mozregression/bisector.py
+++ b/mozregression/bisector.py
@@ -11,7 +11,7 @@ from mozregression.build_range import range_for_inbounds, range_for_nightlies
 from mozregression.errors import LauncherError, MozRegressionError, \
     GoodBadExpectationError
 from mozregression.history import BisectionHistory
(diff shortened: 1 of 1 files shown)


mozregression/bisector.py
```diff
@@ -245,7 +245,21 @@ def handle_merge(self):
...
+            if get_category(most_recent_push.repo_name) != 'integration' and \
+               len(push.changesets) >= 2:
+                jp2 = JsonPushes("mozilla-inbound")
+                try:
+                    data = jp2.pushes_within_changes(
+                        push.changesets[0]['node'],
+                        push.changesets[-1]['node'])
+                except MozRegressionError:
+                    return
+                LOG.info("************* Switching to mozilla-inbound")
```
wlach: Could you make the log message reflect the fact that we couldn't automatically determine the branch, and we're falling back to inbound? I think that would be helpful for debugging.

PR #427, opened 2016-05-04
Allow for a bisection without a download manager

EricRahm:
This allows a bisection to take place without using a download manager.
This is particularly useful bisecting builds where the data we're
interested in is already in test logs, so no binary is necessary.

My use case was to create a custom TestRunner that greps through logs looking for a specific warning. I use `mozregression.bisector.Bisector` to drive the bisection and let the TestRunner take care of downloading.

@parkouss What do you think?

### DIFF
--- a/mozregression/bisector.py
+++ b/mozregression/bisector.py
@@ -416,7 +416,7 @@ class Bisection(object):
         found, mid_point, build_infos, persist_files = self._find_approx_build(
             mid_point, build_infos
         )
-        if not found:
+        if not found and self.download_manager:
             # else, do the download. Note that nothing will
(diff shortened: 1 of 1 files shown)


parkouss:
I'm longer really active on mozregression unfortunately :(. Though the idea seems good to me, and the implementation seems simple and not to break things (I assume you tested it).

You should discuss that with @wlach I think. :)

EricRahm:
@wlach Eventually I'd like to add a more general solution to running a bisection without downloading, but for now this one line change meets my needs. The travis-ci tests were happy (appveyor failed setting up, not running). Would you mind integrating and cutting a release?

PR #428, opened 2016-06-07
Bug 1278683 - Update config and build info handling

EricRahm:
Taskcluster routes (for linux64 debug at least) have temporarily changed
from using 'debug' to 'dbg'. Eventually this behavior will revert but
for now we need to special case it.

Additionally the binary archive name for builds in taskcluster has
changed from specifying the platform and architecture to just using a
generic term, 'target', for all platforms. The build regex has been
updated to handle that behavior.


EricRahm:
@wlach can you take a look?

wlach:
Thanks!

PR #431, opened 2016-09-07
Bug 1301021 - update taskcluster version for mozregression

Conjuror:
update taskcluster version to newer than 0.3.4

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ DEPENDENCIES = [
     'mozdevice == 0.47',
     # later taskcluster versions seem not to work, see:
     # https://bugzilla.mozilla.org/show_bug.cgi?id=1265498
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -45,7 +45,7 @@ def run_tests(self):
     'mozdevice == 0.47',
     # later taskcluster versions seem not to work, see:
     # https://bugzilla.mozilla.org/show_bug.cgi?id=1265498
-    'taskcluster == 0.0.32',
+    'taskcluster >= 0.3.4',
```
wlach: We do want to pin the version to guard against API breakage (see the bug linked). We can also remove the link.

wlach:
I commited. a slightly modified version of this in d43ec78b34bd48959419c2ad1e6c0266ac0b9446

PR #432, opened 2016-09-09
fall back taskcluster version to 0.0.32 due to API breakage in 0.3

Conjuror:


wlach:
Closing this, as my understanding is that we'll be staying on the more recent version of taskcluster.

PR #434, opened 2016-09-21
Allow longer line lengths

EricRahm:
@wlach Does this seem better? I chose 100 as a reasonable slightly longer length (default is 79).


wlach:
works for me!

PR #435, opened 2016-09-25
Fix bisect progress type mismatch, convert dates to datetimes

jay:
I tried to do a bisect in mozregression recently and it failed with **TypeError: unsupported operand type(s) for -: 'datetime.date' and 'datetime.datetime'**. It traces back to [`_print_progress` in bisector](https://github.com/mozilla/mozregression/blob/1e38c1e/mozregression/bisector.py#L157):
~~~python
    def _print_progress(self, new_data):
        next_good_date = new_data[0].build_date
        next_bad_date = new_data[-1].build_date
        next_days_range = abs((next_bad_date - next_good_date).days)
~~~
~~~
next_good_date: 2016-07-30 03:02:04       <--- this is a datetime
next_bad_date:  2016-08-27                <--- this is a date
~~~
It appears that Python does not allow getting the delta of a date and a datetime. To fix this I am proposing a change that converts both to datetimes, as well as a similar fix for the self good/bad date which I had a similar problem with. Here is what it looks like with the fix:
~~~
next_good_date: 2016-07-30 03:02:04       <--- this is a datetime
next_bad_date:  2016-08-27 00:00:00       <--- this is a datetime
~~~

**I'm not proficient in python** and it may be better to handle this earlier when a build date is created. Though I did find [code in testrunner](https://github.com/mozilla/mozregression/blob/1e38c1e/mozregression/test_runner.py#L28) that checks a build date to see if it's an instance of datetime, so I figured it is expected that build date won't always be a datetime.

Here's the full trace:
~~~
  File "C:\Python27\Scripts\mozregression-script.py", line 9, in <module>
    load_entry_point('mozregression==2.3.8', 'console_scripts', 'mozregression')
()
  File "c:\python27\lib\site-packages\mozregression\main.py", line 308, in main
    sys.exit(method())
  File "c:\python27\lib\site-packages\mozregression\main.py", line 127, in bisec
t_nightlies
    result = self._do_bisect(handler, good_date, bad_date)
  File "c:\python27\lib\site-packages\mozregression\main.py", line 202, in _do_b
isect
    return self.bisector.bisect(handler, good, bad, **kwargs)
  File "c:\python27\lib\site-packages\mozregression\bisector.py", line 585, in b
isect
    return self._bisect(handler, build_range)
  File "c:\python27\lib\site-packages\mozregression\bisector.py", line 656, in _
bisect
    result = bisection.handle_verdict(index, verdict)
  File "c:\python27\lib\site-packages\mozregression\bisector.py", line 548, in h
andle_verdict
    self.handler.build_bad(mid_point, self.build_range)
  File "c:\python27\lib\site-packages\mozregression\bisector.py", line 122, in b
uild_bad
    self._print_progress(new_data)
  File "c:\python27\lib\site-packages\mozregression\bisector.py", line 157, in _
print_progress
    next_days_range = abs((next_bad_date - next_good_date).days)
TypeError: unsupported operand type(s) for -: 'datetime.date' and 'datetime.date
time'
~~~
Also I had some problems using mozregression from the virtualenv, Python kept using supporting scripts from the global install of mozregression. Where is the appropriate place to ask about that in detail?

### DIFF
--- a/mozregression/bisector.py
+++ b/mozregression/bisector.py
@@ -8,6 +8,7 @@ import threading
 from mozlog import get_proxy_logger
 
 from mozregression.build_range import range_for_inbounds, range_for_nightlies
+from mozregression.dates import to_datetime
 from mozregression.errors import LauncherError, MozRegressionError, \
     GoodBadExpectationError, EmptyPushlogError
 from mozregression.history import BisectionHistory
@@ -154,13 +155,15 @@ class NightlyHandler(BisectorHandler):
     def _print_progress(self, new_data):
         next_good_date = new_data[0].build_date
         next_bad_date = new_data[-1].build_date
-        next_days_range = abs((next_bad_date - next_good_date).days)
+        next_days_range = abs((to_datetime(next_bad_date) -
+                               to_datetime(next_good_date)).days)
         LOG.info("Narrowed nightly regression window from"
                  " [%s, %s] (%d days) to [%s, %s] (%d days)"
                  " (~%d steps left)"
                  % (self.good_date,
                     self.bad_date,
-                    abs((self.bad_date - self.good_date).days),
+                    abs((to_datetime(self.bad_date) -
+                         to_datetime(self.good_date)).days),
                     next_good_date,
                     next_bad_date,
                     next_days_range,


jay:
CI had this failing [test_print_progress](https://github.com/mozilla/mozregression/blob/1e38c1e/tests/unit/test_bisector.py#L98-L113) so I changed it to output just the dates even if a datetime is available. 
`Narrowed nightly regression window from [2016-07-30, 2016-09-23] (55 days) to [2016-07-30, 2016-08-27] (27 days) (~4 steps left)`
Alternatively the test could be modified so that this is acceptable:
`Narrowed nightly regression window from [2016-07-30 00:00:00, 2016-09-23 00:00:00] (55 days) to [2016-07-30 00:00:00, 2016-08-27 00:00:00] (27 days) (~4 steps left)`

jay:
A different idea since datetimes don't appear to be used in the function is instead of converting everything to datetime, convert everything to date, see [master...jay:fix_bisect_progress_take2](https://github.com/mozilla/mozregression/compare/master...jay:fix_bisect_progress_take2?expand=1)

wlach:
Hey @jay, thanks for reporting the bug and providing a potential patch. I'm a bit concerned that there's a larger issue in mozregression, since I think at this point in the program both variables should be plain dates. I suspect we should be converting something to Date earlier in the control flow. 

It's issues like this that motivated adding type annotations to python. :)

Could you print a full log of what interactions you made with mozregression to get to this point? Also, what platform are you on?

jay:
> Could you print a full log of what interactions you made with mozregression to get to this point? Also, what platform are you on?

Sure,

Windows 7 x64 Enterprise
Python 2.7.12 x64
mozregression from master ([gui-0.9.4](https://github.com/mozilla/mozregression/tree/gui-0.9.4))

I was trying to identify when Firefox's popup menus started appearing on the wrong screen for me (since filed as Mozilla bug [1305278](https://bugzilla.mozilla.org/show_bug.cgi?id=1305278)). I took the build ID from a bad version and went back about 2 months and found a good version, and took that build ID as well. I passed both to mozregression:
~~~
setup.py install && mozregression --bad 20160923030450 --good 20160730030204
~~~
Then I marked commits in this order: good, bad, bad.
~~~
INFO: Running mozilla-central build for buildid 20160730030204   <--- good
INFO: Running mozilla-central build for buildid 20160923030450   <--- bad
INFO: Running mozilla-central build for 2016-08-27               <--- bad
~~~
That last bad is the point where it errors with the trace I posted, because it can't get the delta
~~~
next_good_date: 2016-07-30 03:02:04       <--- this is a datetime
next_bad_date:  2016-08-27                <--- this is a date
~~~
With that fixed the same error not surprisingly will later happen in `self.bad_date - self.good_date` so that should be considered as well.

wlach:
Ok, thanks! It took a bit of unwinding, but I figured out that it's possible for us to be fetching nightlies either with buildid (which are a datetime) or a date passed from the command line. I think most of the time when you get to the function you asserted on, you'd only be dealing with datetime, which is how this bug slipped through.

I think the easiest solution is to use the utility function to_datetime to convert the values when calling abs((current_bad_date - current_good_date).days). Do you want to give that a shot? Should be a one-liner.

https://github.com/mozilla/mozregression/blob/master/mozregression/dates.py#L31

jay:
> I think the easiest solution is to use the utility function to_datetime to convert the values when calling abs((current_bad_date - current_good_date).days). Do you want to give that a shot? Should be a one-liner.

Thanks for the feedback William. I've amended the commit associated with this PR with the suggested changes.

wlach:
Thanks Jay! That looks great. I will test it out tomorrow and merge if everything looks ok.

Regarding your virtualenv question (sorry I missed it earlier), I am not sure if I know the answer as I am not familiar with the Windows environment. :(

jay:
TL;DR: It's more some general issues about virtualenv and mozregression. I am using virtualenv-win which is a wrapper for virtualenv but I'd guess the same question could apply to other operating systems. I'm hesitant to create the noise but I will ask here and even if you don't know the answer maybe you can tell me where is an appropriate place to ask.

---

I followed [the GUI directions](https://github.com/mozilla/mozregression/blob/master/gui/README.rst) to set up a virtual environment for developing mozregression in Windows, which is basically this:
~~~
curl --fail -OL http://sourceforge.net/projects/pyqt/files/PyQt4/PyQt-4.11.4/PyQt4-4.11.4-gpl-Py2.7-Qt4.8.7-x64.exe
echo "1ad2fd94ec8f6e8287d6ae80d520b4974b90ca30dbeae4256f3317ce4d79bc23 *PyQt4-4.11.4-gpl-Py2.7-Qt4.8.7-x64.exe" | sha256sum -c
# if the checksum is OK then install
PyQt4-4.11.4-gpl-Py2.7-Qt4.8.7-x64.exe
pip install pytest flake8 virtualenvwrapper-win
mkvirtualenv --system-site-packages -p C:\Python27\python.exe mozregression
# mkvirtualenv switches to the mozregression env after creating it
pip install -r requirements-gui-dev.txt
~~~

At the time I already had mozregression pip installed system-wide so when I tried to run mozregression from the virtual environment it was using some files from the system-wide installation. In other words I could run mozregression in the virtual environment and some changes I made in the repo were being used but the other changes I made (eg in bisector) were not. What I do not understand is if that is expected behavior?

To remedy I uninstalled system-wide mozregression, and each time I want to run it in the repo I run it from the virtual environment like `setup.py install && mozregression opts`. Is that the correct way to do it, or can I run it directly from the repo so that I don't have to reinstall it each time I make a change?

Neither of the documented ways ([1](https://github.com/mozilla/mozregression/blob/master/README.md#building-and-developing-mozregression), [2](https://github.com/mozilla/mozregression/blob/master/gui/README.rst#launching-the-application)) say how to launch the command-line mozregression in the virtual environment or discuss the possibility of a system-wide conflict.

wlach:
@jay I think normally the virtualenv-provided packages would be preferred, so I'm not sure why that would be happening. Normally I do a "local" install of mozregression when working on it, so I don't have to move/copy anything anywhere during development. You should be able to do this with:

```sh
pip install -e /path/to/mozregression-checkout
```

In any case, thanks again for the patch. It works great!

PR #437, opened 2017-01-16
Downgrade flake8 to version 2.5.4 and peg various python linting requirements

wlach:
Later versions seem to be detecting new "problems" which we may
want to fix at some point, but not now.


wlach:
This is still failing travis, but less so than before.  :)

PR #438, opened 2017-01-17
PEP8

stepshal:


wlach:
@stepshal Thanks! This is awesome, and right on schedule. :) In #437 I pegged the version of flake8 (and its dependencies) to an older version. I think we want to keep the pegging, but with your patch, we can now upgrade to the latest version -- do you think you could update your PR here to upgrade the version of flake8 (and its dependencies)? #437 should tell you which files need to be modified.

stepshal:
@wlach Done.

wlach:
Thank you!

PR #440, opened 2017-02-06
Bug 1334669 - Don't show bug numbers if >1 push that could be bad

wlach:
We weren't checking that there wasn't any pushes between the good/bad
results in inbound bisection, which would cast doubt that the first
found "bad" push really was the bad one.

### DIFF
--- a/mozregression/main.py
+++ b/mozregression/main.py
@@ -31,6 +31,7 @@ from mozregression.download_manager import BuildDownloadManager
 from mozregression.persist_limit import PersistLimit
 from mozregression.fetch_build_info import (NightlyInfoFetcher,
                                             InboundInfoFetcher)
+from mozregression.json_pushes import JsonPushes
 from mozregression.bugzilla import find_bugids_in_push, bug_url
 from mozregression.approx_persist import ApproxPersistChooser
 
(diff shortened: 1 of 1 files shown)


mozregression/main.py
```diff
@@ -160,32 +161,41 @@ def _bisect_inbounds(self, good_rev, bad_rev, ensure_good_and_bad=False,
...
+                        bugids = find_bugids_in_push(
+                            handler.build_range[1].repo_name,
+                            handler.build_range[1].changeset
+                        )
+                        if len(bugids) == 1:
+                            word = 'fix' if handler.find_fix else 'regression'
+                            LOG.info("Looks like the following bug has the "
+                                     " changes which introduced the"
+                                     " {}:\n{}".format(word,
+                                                       bug_url(bugids[0])))
```
zapion: Here's an unclosed if block, might not be a big deal since it works before.

mozregression/main.py
```diff
@@ -160,32 +161,41 @@ def _bisect_inbounds(self, good_rev, bad_rev, ensure_good_and_bad=False,
                                  ensure_good_and_bad=ensure_good_and_bad)
         result = self._do_bisect(handler, good_rev, bad_rev, expand=expand)
         if result == Bisection.FINISHED:
-            LOG.info("Oh noes, no (more) inbound revisions :(")
```
wlach: Hmm. I think we should probably have the log message, now that I think about it. Probably a more meaningful one though -- this isn't an error condition, as is currently implied. I'll revert this change and fix things in a followup.

PR #441, opened 2017-02-09
Bug 1335692 - Bisecting nightlies given firefox releases number doesn…

zapion:
- Update firefox 53 releases
- Add warning log for non-supported release index.


mozregression/cli.py
```diff
@@ -392,7 +392,9 @@ def _convert_to_bisect_arg(self, value):
                                  % (new_value, value))
                 value = new_value
             except UnavailableRelease:
-                pass
+                self.logger.warning(
+                    "Target release version %s is not yet supported." % (value)
+                )
```
wlach: Two things:

1. Could you change this from warning to error?
2. Could you add a "raise" statement here? This will make mozregression error out (which is what we want)

mozregression/cli.py
```diff
@@ -386,13 +386,10 @@ def _convert_to_bisect_arg(self, value):
...
             value = parse_date(value)
         except DateFormatError:
-            try:
-                new_value = parse_date(date_of_release(value))
-                self.logger.info("Using date %s for release %s"
-                                 % (new_value, value))
-                value = new_value
-            except UnavailableRelease:
-                pass
+            new_value = parse_date(date_of_release(value))
```
zapion: @wlach I just realized we shouldn't catch anything here.

wlach: I think it's good to log the message here, as it's a bit more friendly than just throwing the exception.

zapion: @wlach https://github.com/mozilla/mozregression/commit/e6c8c3afde4f388d34c1b516db52f51c01ca9f79#diff-a753870fc2c6fa4864480ab7b1c1cd57L479
I checked last commit of these, this function seems to return only datetime.  But it might be hash which leads to match inbound builds.  Therefore we passed UnavailableRelease here.
Since this change broke 8 unittests we should be careful.

wlach: Ah I see, yes, we'll need to revise the patch then. I think it is probably best to just leave things as-is, but add some more logging information about mozregression is doing: whether it is bisecting by date, commit hash, or release number. Could you add that in?

zapion: I think that makes sense, but probably slightly adjustment for the flow will be nice in a following bug. will revise the patch with more logs.

mozregression/cli.py
```diff
@@ -392,7 +392,7 @@ def _convert_to_bisect_arg(self, value):
                                  % (new_value, value))
                 value = new_value
             except UnavailableRelease:
-                pass
+                self.logger.info("Using hash to bisect inbound")
```
zapion: @wlach , leave log for switching to inbound, hope it doesn't look annoying.

wlach:
I modified the log message a bit, but this looks good. Thank you!

PR #443, opened 2017-02-24
Bug 1342405 - Add date for release 53

staktrace:


wlach:
We actually added this in #441, but it makes more sense to do this as a seperate commit, so I'll go ahead and merge. Thank you!

PR #447, opened 2017-04-12
Bug 1334815 - Bisecting Fennec with taskcluster builds fails

zapion:
taskcluster changed artifact name so we can't match build file and download it.
I actually don't know which build info file (there're two, target.txt, target_info.txt) is correct.  please help to clarify, thank.


zapion:
Updated per JanH's feedback. @wlach , can you help to review? thanks!

wlach:
Looks fine, I will merge!

PR #453, opened 2017-06-12
Bug 1348867

KSFT:


wlach:
Thanks for this! There are a few minor bugs I noticed, hopefully they won't be too hard to resolve:

1. If you select another branch name, go to the screen to select between date/release, then go back and clear the branch name, the "release" field is still disabled (it should be enabled, since no branch implies mozilla-central).
2. If I keep the default, then go to the screen to select between date/release, select a release, then go back to the screen to the first screen and select a branch other than m-c, when I go back to the date/release screen it's still selecting a release (I would have expected it to revert to a default)

wlach:
Merged manually in 6389a47c102d184f025f350b686e1f55c25559df

PR #454, opened 2017-06-16
Stop blaming a particular bug, because the code that comes up with th…

staktrace:
…e bug to blame is broken, and this just misleads people.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1373664 for an example.

### DIFF
--- a/mozregression/main.py
+++ b/mozregression/main.py
@@ -178,6 +178,10 @@ class Application(object):
                     return self._bisect_inbounds(good_rev, bad_rev,
                                                  expand=DEFAULT_EXPAND)
                 else:
+                    # This code is broken, it prints out the message even when
+                    # there are multiple bug numbers or commits in the range.
+                    # Somebody should fix it before re-enabling it.
+                    return 0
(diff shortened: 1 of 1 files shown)


mozregression/main.py
```diff
@@ -178,25 +178,29 @@ def _bisect_inbounds(self, good_rev, bad_rev, ensure_good_and_bad=False,
                     return self._bisect_inbounds(good_rev, bad_rev,
                                                  expand=DEFAULT_EXPAND)
                 else:
+                    # This code is broken, it prints out the message even when
+                    # there are multiple bug numbers or commits in the range.
+                    # Somebody should fix it before re-enabling it.
+                    pass
```
wlach: Could you change this to just 'return 0' and remove the changes below? I wouldn't make a big issue of it but the way you wrote the patch is causing travis to fail.

staktrace:
Yup, patch updated. (Sorry I don't do python much).

wlach:
Thanks!

PR #456, opened 2017-06-23
Use requests[security], pinned to 2.15.1

wlach:
Fixes: https://bugzilla.mozilla.org/show_bug.cgi?id=1375832


wlach:
Latest version of requests doesn't seem to install correctly in the travis environment and I honestly don't care enough to fix it. Let's pin to 2.15.1, the last update before it sounds like they did a major refactoring of dependencies.

PR #459, opened 2017-08-21
See what happens if we depend/include python-adb

wlach:


wlach:
... that was a disappointing experiment. :(

I think the way forward is to include an adb executable with the gui version of mozregression. Closing this for now.

PR #461, opened 2017-09-05
Bump version of cx_freeze

wlach:


wlach:
Looks like the latest version of cx_freeze is broken for older versions of python:

https://github.com/anthony-tuininga/cx_Freeze/pull/272

Let's see if 5.0.1 works, it should.

wlach:
5.0.1 builds successfully but doesn't work on windows due to: https://bugzilla.mozilla.org/show_bug.cgi?id=1332162

arrrrrgh

wlach:
Ok, this is too hard. Let's revert to precise, which will fix travis on linux. We can upgrade cx_freeze along with everything in some mythical future where everything is using python3.

PR #466, opened 2017-10-04
Get version from scm

wlach:


wlach:
Seems to mostly work, except the gui says that the version is None in "about".

PR #467, opened 2017-10-04
updating README.md file

tiagomoraismorgado88:
updating README.md file


wlach:
Thank you, these are reasonable changes (except for the mozregression capitalization changes, where I prefer to be consistently lower-case)

PR #468, opened 2017-10-13
Encode environment variables as ascii when launching mozregression

wlach:
It seems in certain cases, we were trying to pass unicode values to
the process runner, which resulted in an error.

See: https://github.com/mozilla/geckodriver/issues/972#issuecomment-336475663


mozregression/test_runner.py
```diff
@@ -196,7 +196,7 @@ def evaluate(self, build_info, allow_back=False):
 
             env = dict(os.environ)
             for k, v in variables.iteritems():
-                env['MOZREGRESSION_' + k.upper()] = v
+                env['MOZREGRESSION_' + k.upper()] = v.encode('ascii', errors='ignore')
```
whimboo: In line 192 there is already a call to `str()`. I think that one should be removed in favor of the conversion you just added.

wlach: You are right, it looks like the error we're seeing must have some other root cause. :( Since if the value was unicode we would have seen an exception

whimboo: The only value we do not explicitly convert into a string is the binary. Could we have a strange value here?

wlach: Seems unlikely but possible I suppose?

wlach: Maybe we can just do what you suggest and hope for the best. :)

PR #472, opened 2017-11-01
Fix typo in message about 64-bit systems

dholbert:
Right now the message about default build type (64-bit vs. 32-bit) incorrectly refers to them as "64 bits" and "32 bits".  The correct/common adjective for these types of builds are "64-bit" vs. "32-bit" (singular bits, and with a hyphen).  See e.g. https://support.mozilla.org/en-US/kb/switch-32-bit-64-bit as one reference on this language.


wlach:
thank you!

PR #479, opened 2017-12-22
Bug 1426105 - Make elimination branch detection work with find fix runs.

KwanEsq:
Call _reverse_if_find_fix() when there is no explicit branch in commit message,
so find fix runs can continue onto integration branches.


KwanEsq:
Feel free to fiddle with the commit message, I couldn't come up with one I really liked :/

wlach:
Hey! Thank you very much for this, I will look at it in the new year.

wlach:
This looks fine! Sorry for the delay in reviewing. I will release a new version shortly with these changes.

PR #486, opened 2018-04-11
Handle transposed repo names in merge commit messages

KwanEsq:
Some merge commit messages have the positions of the originating and target repos reversed.  This breaks mozregression's branch detection since it assumes the originating repo is always in the first position.  If the first repo is "mozilla-central" assume this is a transposed message and take the second repo's name.

I put the logging in a separate commit in case it wasn't wanted.

https://bugzilla.mozilla.org/show_bug.cgi?id=1450986

### DIFF
--- a/mozregression/bisector.py
+++ b/mozregression/bisector.py
@@ -265,7 +265,7 @@ class InboundHandler(BisectorHandler):
         push = jp.push(most_recent_push.changeset, full='1')
         msg = push.changeset['desc']
         LOG.debug("Found commit message:\n%s\n" % msg)
-        branch = find_branch_in_merge_commit(msg)
+        branch = find_branch_in_merge_commit(msg, most_recent_push.repo_name)
         if not (branch and len(push.changesets) >= 2):
             # We did not find a branch, lets check the integration branches if we are bisecting m-c
             if get_name(most_recent_push.repo_name) == 'mozilla-central' and \
--- a/mozregression/branches.py
+++ b/mozregression/branches.py
@@ -86,16 +86,22 @@ get_branches = BRANCHES.get_branches
 get_category = BRANCHES.get_category
 
 
-RE_MERGE_BRANCH = re.compile(r"merge ([\w-]+) to [\w-]+.*", re.I)
+RE_MERGE_BRANCH = re.compile(r"merge ([\w-]+) to ([\w-]+).*", re.I)
 
 
-def find_branch_in_merge_commit(message):
+def find_branch_in_merge_commit(message, current_branch):
     """
     Try to extract the branch name where commits comes from in a merge commit
     message.
+
+    Some commit messages incorrectly transpose the repo names.  If the first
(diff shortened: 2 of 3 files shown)


wlach:
@KwanEsq you are the best! thank you so much. There is a very minor flake8 error in your commit (see travis):

```
Running: |flake8 --max-line-length=100 mozregression tests setup.py ./check.py|
mozregression/branches.py:14:1: E302 expected 2 blank lines, found 1
  File "./check.py", line 48, in <module>
    run(['flake8', '--max-line-length=100', 'mozregression', 'tests', 'setup.py', __file__])
  File "./check.py", line 32, in run
    check_call(cmd, **kwargs)
  File "/usr/lib/python2.7/subprocess.py", line 511, in check_call
    raise CalledProcessError(retcode, cmd)
```

Could you fix that and update the PR?

wlach:
Actually I can probably just fix this myself, nevermind...

KwanEsq:
Hmm, I think I've fixed that but confused the CI by force-pushing an amended commit instead of adding a new one.  Certainly the "Files changed" view shows the newline under `LOG = get_proxy_logger('Branches')` that I was missing in the first version.

KwanEsq:
Ah, I have broken a test however.

KwanEsq:
Hmm, so something I'd thought about then forgot to check, is merge handling ever used anywhere but m-c -> m-i/autoland?  If so my current approach won't work for that, since it assumes it.
I guess if it is used elsewhere `find_branch_in_merge_commit` could return both branch names (and I guess be renamed to `find_branches_[...]`), and bisector.py could do the swap itself only if `most_recent_push.repo_name == originRepo` or something.

If the current approach is fine I'd guess the test just needs its expectation updating.

wlach:
@KwanEsq it looks like the function `find_branch_in_merge_commit` is called only from `handle_merge`, here:

https://github.com/mozilla/mozregression/blob/5b468a469280978bac6401db5d772e5973e46135/mozregression/bisector.py#L268

Could you add a parameter called current_branch to `find_branch_in_merge_commit` and then do the transposition on the destination repo matching that? You should be able to get the current branch easily inside the handle_merge function. The end results should be both more general and more robust than what it is there now. I don't think it is common (or even possible) to go from bisecting an inbound branch to central, but you never know.

KwanEsq:
@wlach done, though I'm not sure about the test updates, if they should all have explicit currents or not.
Can fold the new commit into the first one once we're happy with it.
Apologies for not running the tests locally before originally opening the PR.

(also it seems I was right about my earlier force-pushed amend confusing CI)

wlach:
@KwanEsq I think it would be best to give the test explicit currents, just pick the most rationale option given the input to the test function. Thanks again for doing this work, it's really appreciated.

KwanEsq:
Sweet, all green so I'm going to fold the changes into the first commit.

PR #487, opened 2018-05-02
Disable what's new page

marco-c:


wlach:
Anything particularly annoying about this preference @marco-c? I'd prefer to leave the settings as default unless there's a compelling reason to change them.

marco-c:
The what's new page is shown whenever you run a Firefox version that is different from the one you previously run, so it's basically very often loaded when running mozregression.
This makes it hard to reproduce bugs in session restore.

wlach:
Makes sense, I will merge the patch! Thanks.

Darkspirit:
I just wanted to let you know that you broke my **stupid** workflow with a recent mozregression update. Patches have consequences. :D I can't explain why I stuck with it.
All my old mozregression STRs are [broken](https://bugzilla.mozilla.org/show_bug.cgi?id=1401665#c10) on BMO, but out of distress I [switched to -a](https://bugzilla.mozilla.org/show_bug.cgi?id=1478339#c2).
You forced me to improve myself. Although it's a bit sad that old STRs are broken..^^

marco-c:
Sorry about that :(
We could set `browser.startup.homepage_override.mstone` when `startup.homepage_welcome_url` is not the default, but I'm not sure it's worth it.

PR #488, opened 2018-06-10
Bug 1128240 - Fetch merge dates for Firefox versions from hg tags.

KwanEsq:
Seeing as creating a new mozregression manually for each new Firefox version seemed a bit tedious I thought I'd give this a go, looking for some feedback @wlach .

(Don't need to delete as many/any of the existing dates that I have, but I wanted to test it with some different version numbers)

Apologies if some of this isn't particularly idiomatic python.
I'm also not sure if the date handling will do the right thing when run in other timezones, but it might not be a huge deal if it downloads a version one day out.

### DIFF
--- a/mozregression/releases.py
+++ b/mozregression/releases.py
@@ -1,4 +1,8 @@
+import re
+
+from datetime import date
 from mozregression.errors import UnavailableRelease
+from mozregression.network import retry_get
 
 
 def releases():
@@ -11,7 +15,7 @@ def releases():
     # using the ones in the "beta" column (formerly "aurora"). This is because
     # the merge date for beta corresponds to the last nightly for that
     # release. See bug 996812.
-    return {
+    releases = {
         5: "2011-04-12",
         6: "2011-05-24",
         7: "2011-07-05",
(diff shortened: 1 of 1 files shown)


mozregression/releases.py
```diff
@@ -63,13 +67,36 @@ def releases():
...
-        57: "2017-09-21",
-        58: "2017-11-13",
-        59: "2018-01-22",
-        60: "2018-03-12"
+        56: "2017-08-02"
     }
 
+    def filter_tags(tag_node):
+        RE_TAG = re.compile("^FIREFOX_NIGHTLY_(\d+)_END$")
+        match = RE_TAG.match(tag_node["tag"])
```
wlach: Probably no need to define a once-off regular expression here. Just:

```py
match = re.match("^FIREFOX_NIGHTLY_(\d+)_END$", tag_node["tag"])
```

should suffice.

mozregression/releases.py
```diff
@@ -63,13 +67,36 @@ def releases():
...
+        return int(match.group(1)) > 56 if match else False
+
+    def map_tags(tag_node):
+        RE_VER = re.compile("_(\d+)_")
```
wlach: Same comment as above.

KwanEsq:
@wlach Thanks! Done.
(I'm still not sure if it's better on here to push the fixups as a separate commit, and amend later (or let the reviewer/pusher do so))

Also I didn't mention it before but just as an historical note, for 58 and 59 this ended up being a day different than the previous hard-coded values.
58: 2017-11-13 -> 2017-11-12
59: 2018-01-22 -> 2018-01-21

wlach:
@KwanEsq Either is fine for a small PR like this, though I personally prefer an additional commit just because it makes it easier to see exactly what's changed. Thanks so much! This should take some of the grunt work out of updating mozregression as new releases get merged into beta.

KwanEsq:
\o/ Thank you!

PR #489, opened 2018-06-11
gui: fixup signal connection in wizards

xrmx:
Fixes the following backtrace when running the bisection or single build wizards:

```
  File "/home/rm/src/mozregression/gui/mozregui/mainwindow.py", line 121, in single_run
    self._start_runner(SingleRunWizard, self.single_runner)
  File "/home/rm/src/mozregression/gui/mozregui/mainwindow.py", line 99, in _start_runner
    wizard = wizard_class(self)
  File "/home/rm/src/mozregression/gui/mozregui/wizard.py", line 309, in __init__
    parent=parent)
  File "/home/rm/src/mozregression/gui/mozregui/wizard.py", line 257, in __init__
    "currentIndexChanged")
TypeError: setDefaultProperty(self, str, str, QT_SIGNAL): argument 3 has unexpected type 'str'
```

### DIFF
--- a/gui/mozregui/wizard.py
+++ b/gui/mozregui/wizard.py
@@ -2,7 +2,7 @@ import mozinfo
 import datetime
 from PyQt4.QtGui import (QWizard, QWizardPage, QStringListModel, QMessageBox,
                          QCompleter, QApplication)
-from PyQt4.QtCore import QString, QDate, pyqtSlot as Slot, Qt
+from PyQt4.QtCore import QString, QDate, pyqtSlot as Slot, Qt, SIGNAL
 
 from ui.intro import Ui_Intro
 from ui.build_selection import Ui_BuildSelectionPage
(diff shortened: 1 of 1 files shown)


wlach:
@xrmx sorry for the late reply on this -- this patch looks totally reasonable, but can you give some steps to reproduce the problem that motivated you to write it? I can't actually generate a backtrace with the way it currently works.

xrmx:
How to reproduce:
```
python gui/build.py run
click on File -> Run a new bisection
```

and I get a Crash Dialog

wlach:
ok, I can't reproduce this under Windows but I totally believe this could be a problem under some versions of Linux. I'll give it a try when I have access to such a machine later this week and merge when I do so. Thanks for the patch!

wlach:
finally had a chance to verify this on a linux machine. does indeed seem to be a problem, strangely only in just this environment. thanks for the patch!

PR #501, opened 2018-11-14
Bug 1507225 - Add basic support for launching the GeckoView Example app

staktrace:


mozregression/launchers.py
```diff
@@ -388,6 +388,72 @@ def get_app_info(self):
...
+
+    def _stop(self):
+        self.adb.stop_application(self.package_name)
+        if self.adb.exists(self.remote_profile):
+            self.adb.rm(self.remote_profile, recursive=True)
+
+    def get_app_info(self):
+        return self.app_info
+
+
```
wlach: It looks like this is 99% the same as the FennecLauncher class elsewhere in the file. Could you create a base class called `AndroidLauncher` and have both this and FennecLauncher inherit from it?

mozregression/fetch_build_info.py
```diff
@@ -270,6 +271,7 @@ def find_build_info(self, date, fetch_txt_info=True, max_workers=2):
             for thread in threads:
                 while thread.is_alive():
                     thread.join(0.1)
+            LOG.debug("got valid_builds %s" % valid_builds)
```
wlach: Could you remove these debugging logs as well?

staktrace: Is there rationale for when debug logs are ok vs not? I found these logs useful when debugging the code, so I figured they might be useful to others too.

mozregression/launchers.py
staktrace: Yup, will do

mozregression/fetch_build_info.py
wlach: Not really. Actually, on second thought, let's leave these in. They shouldn't be visible under normal circumstances.

wlach:
Thank you! I will do a new release with these changes soon

PR #503, opened 2018-11-15
Don't show debugging messages in GUI

wlach:


wlach:
Somewhat on the fence on whether I should add an option for this, but leaning towards "not worth the effort" (in all the years I've maintained mozregression, I haven't needed the debugging logs for anything).

PR #504, opened 2018-11-20
Warn on using = instead of : in pref argument

staktrace:
I do this all the time and the silent-drop behaviour of mozregression is kinda annoying

### DIFF
--- a/mozregression/cli.py
+++ b/mozregression/cli.py
@@ -293,7 +293,7 @@ def parse_bits(option_bits):
         return mozinfo.bits
 
 
-def preferences(prefs_files, prefs_args):
+def preferences(prefs_files, prefs_args, logger):
     """
     profile preferences
     """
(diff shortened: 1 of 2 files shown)


wlach:
Note also the unit tests are failing (didn't dig into why, aside from the fact that it doesn't seem to be intermittent issue)

staktrace:
I was thinking about allowing both, but it seems like it would be more complex because we need to find whichever one comes first. And what if there are multiple prefs with different separators, do we want to allow only a single separator (either ':' or '=') for all the prefs, or allow each pref to have a separate one? I assume pref values can have these chars so the number of cases to handle increases dramatically if we want to support both.

I did fix the test failure though ^

wlach:
Ok, makes sense, let's just keep things simple for now and just print the warning. I don't think preferences should have either character in them, but I don't know for certain that would never happen and I guess we're just doubling our chances of problems if we allow both.

PR #510, opened 2019-01-18
Bug 1464127: Bypass profile downgrade detection when using the same profile for bisection.

Mossop:
Once Bug 1455707 lands (planned for Jan 30th), running an older build on a profile is going to essentially bail out. This passes the necessary command line argument to bypass that in the case where every test run uses the same profile.

I'm afraid I haven't been able to run the GUI tests because they don't seem to work for me for some reason.

### DIFF
--- a/gui/mozregui/build_runner.py
+++ b/gui/mozregui/build_runner.py
@@ -152,6 +152,9 @@ class AbstractBuildRunner(QObject):
         launcher_kwargs['addons'] = options['addons']
         self.test_runner.launcher_kwargs = launcher_kwargs
 
+        if options.profile_persistence in ('clone-first', 'reuse') or options.profile:
+            launcher_kwargs['cmdargs'] = launcher_kwargs['cmdargs'] + ['--allow-downgrade']
+
         self.worker = self.worker_class(fetch_config, self.test_runner,
                                         self.download_manager)
         # Move self.bisector in the thread. This will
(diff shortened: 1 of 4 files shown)


wlach:
I think we need to fix the gui tests before this lands. If you have the patience/time, the instructions to run them are here:

https://github.com/mozilla/mozregression/blob/master/gui/README.rst

Otherwise I can take a look at this when I return from PTO next week (I don't think I'll have time today).

Mossop:
I followed those instructuions and just got an error:

    Running: |python build.py test| in gui
    ERROR: No module named PyQt4.uic
      File "check.py", line 62, in <module>
        run(test_run_cmd + ['build.py', 'test'], cwd='gui')
      File "check.py", line 32, in run
        check_call(cmd, **kwargs)
      File "/usr/local/Cellar/python@2/2.7.15_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 190, in check_call
        raise CalledProcessError(retcode, cmd)
    subprocess.CalledProcessError: Command '['python', 'build.py', 'test']' returned non-zero exit status 1

wlach:
Now seeing some issues with the tests themselves, which I think revealed an issue with the updates to the gui (we were expecting options to be an object as it is on the command line, but in the gui it's a dictionary). Trying another commit.

wlach:
ok that seems to pass the tests at least. I will try to verify that the gui works as expected with the change, then land this if so. thanks @Mossop for doing most of the work here!

wlach:
yep, gui still works. rolling a new release with this pr included

PR #512, opened 2019-01-30
WIP add travis support for macos

jrmuizel:


jrmuizel:
MacOS is failing with
```
ERROR: No module named PyQt4.uic
```

wlach:
I am not entirely sure how to debug this, I don't remember seeing this exact problem but have found pyqt4 to be notoriously difficult to install/configure reliably in general. 

I had been wanting to switch to pyqt5 (which can be installed via the standard python `pip` mechanisms) for exactly this reason, but IIRC that requires upgrading to python 3 for some reason, which we're still a ways from being able to do:

https://bugzilla.mozilla.org/show_bug.cgi?id=1425102

jrmuizel:
So installing pyqt4 from https://github.com/cartr/homebrew-qt4 gets us much closer. Now it's failing with:
```
copying /usr/local/opt/pyqt@4/lib/python2.7/site-packages/PyQt4/libQtCore.dylib -> dist/lib/PyQt4/libQtCore.dylib
```
It tries to copy this file because `otool -L /usr/local/Cellar/python@2/2.7.15_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/PyQt4/QtCore.so` says that it's dependency. However that file doesn't exist and doesn't seem to be needed.

We could exclude this file using binExcludes option but that's not exposed to the command line cxfreeze program. We'd need to switch to using a setup script.

wlach:
Work in progress on python 3 port in #513, which should unlock pip-installable pyqt5 and make this easier.

wlach:
This work is more or less obsoleted by #562. I will file a new PR with Mac support soon.

PR #513, opened 2019-02-22
WIP: Support python 3

wlach:


wlach:
This seems to be failing because we override `raw_input` in some unit tests, which doesn't exist in python 3. We use six's input override as a compatibility shim, probably quite easily fixed.

wlach:
Some of this has already landed, the only relevant part is converting the code using modernize, which I think I'm going to do a little more incrementally (core mozregression first, gui second). Closing this PR

PR #514, opened 2019-03-05
Refactor and fix up logic to find last known good changeset when changing branches from a merge.

karlt:
https://bugzilla.mozilla.org/show_bug.cgi?id=1464937


wlach:
This looks good, thank you. I have tested it locally and it seems to work ok. I'm going to go ahead and merge this. At some point we should really write some proper unit tests to make sure all this logic works as we say/think it does...

PR #515, opened 2019-03-08
Bug 1533517 - Allow the user to specify a path for profile storage on android

staktrace:
On some devices apparently the first-attempted path of
/storage/sdcard0/tests works to create the profile, but then the profile
is not readable from the app being tested (e.g. GVE). This patch allows
the user to override the default path to one that may work better.

### DIFF
--- a/mozregression/cli.py
+++ b/mozregression/cli.py
@@ -141,6 +141,16 @@ def create_parser(defaults):
                         metavar="PATH",
                         help="profile to use with nightlies.")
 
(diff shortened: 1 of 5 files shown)


staktrace:
This might still be useful to land, but we have a different fix in bug 1533517, so I'll close this for now.

wlach:
@staktrace I'm happy to land this, as I think it will be useful in the future. Are there any caveats?

staktrace:
Nope, feel free to merge if you're happy with the patch.

PR #518, opened 2019-03-19
Bug 1536013 - Select win64-aarch64 builds when appropriate.

KwanEsq:
Doing this to hopefully get a build for testing.

### DIFF
--- a/gui/mozregui/wizard.py
+++ b/gui/mozregui/wizard.py
@@ -115,7 +115,8 @@ class IntroPage(WizardPage):
         app_name = str(self.ui.app_combo.currentText())
         bits = int(self.ui.bits_combo.currentText())
 
-        self.fetch_config = create_config(app_name, mozinfo.os, bits)
+        self.fetch_config = create_config(app_name, mozinfo.os, bits,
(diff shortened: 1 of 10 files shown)


KwanEsq:
@wlach ah, I was actually thinking that the proper way to do this would be to update mozinfo and push a new version of that first.  It should probably be returning 64 bits for ARM64, as it does for x86-64 even when running on 32-bit Python.  (It might also want to return aarch64 for processor instead of ARM64, ala x86_64 instead of AMD64)

KwanEsq:
Thanks a lot for the collaborator add!
Now I just need to be careful I don't hit any new buttons...

wlach:
@KwanEsq oh ok, if there's a more elegant fix involving mozinfo, let's go for that. I probably shouldn't review a patch for that, but can help with pushing that out to pypi once it's landed.

wlach:
for now i'll do a new release of mozregression with the other fixes that are pending

KwanEsq:
Whoops, pushed the wrong branch, didn't mean to do that yet.

PR #520, opened 2019-04-02
Shippable

KwanEsq:
Let's get a -gui build for testing.
This also has the win64-aarch64 commits in.

### DIFF
--- a/mozregression/fetch_configs.py
+++ b/mozregression/fetch_configs.py
@@ -96,14 +96,34 @@ class CommonConfig(object):
     Define the configuration for both nightly and inbound fetching.
     """
     BUILD_TYPES = ('opt',)  # only opt allowed by default
+    BUILD_TYPE_FALLBACKS = {}
     app_name = None
 
     def __init__(self, os, bits, processor):
         self.os = os
         self.bits = bits
         self.processor = processor
-        self.build_type = 'opt'
         self.repo = None
+        self.set_build_type('opt')
+        self._used_build_index = 0
+
+    @property
+    def build_type(self):
+        """
+        Returns the currently selected build type, which can change if there
+        are fallbacks specified.
+        """
+        return self.build_types[self._used_build_index]
+
+    def _inc_used_build(self):
+        """
+        Increments the index into the build_types indicating the currently
+        selected build type.
+        """
+        self._used_build_index = (
+            # Need to be careful not to overflow the list
+            (self._used_build_index + 1) % len(self.build_types)
+        )
 
     def build_regex(self):
         """
@@ -161,7 +181,10 @@ class CommonConfig(object):
         :raises: MozRegressionError on error.
         """
         if build_type in self.available_build_types():
-            self.build_type = build_type
+            fallbacks = self.BUILD_TYPE_FALLBACKS.get(build_type)
+            self.build_types = (
+                (build_type,) + fallbacks if fallbacks else (build_type,)
+            )
             return
         raise errors.MozRegressionError(
             "Unable to find a suitable build type %r." % str(build_type)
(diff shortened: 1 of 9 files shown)


mozregression/fetch_configs.py
```diff
@@ -384,38 +402,49 @@ def _common_tk_part(inbound_conf):
 
 
 class FirefoxInboundConfigMixin(InboundConfigMixin):
-    def tk_inbound_route(self, push):
+    def tk_inbound_routes(self, push):
         if self.inbound_branch == 'try' or \
            push.timestamp >= TIMESTAMP_GECKO_V2:
```
wlach: `TIMESTAMP_GECKO_V2` is in 2015, so I think we can just simplify this chunk to make it unconditional, and remove the buildbot fallback below.

mozregression/fetch_configs.py
```diff
@@ -91,24 +91,39 @@ def get_build_regex(name, os, bits, processor, psuffix='', with_ext=True):
...
+        return self.build_types[self._used_build_index]
+
+    def inc_used_build(self):
+        """
+        Increments the index into the build_types indicating the currently
+        selected build type.
+        """
+        self._used_build_index = (
+            # Need to be careful not to overflow the list
+            (self._used_build_index + 1) % len(self.build_types)
```
wlach: Should we throw an exception if we overflow? I'm pretty sure that should never happen.

tests/unit/test_fetch_configs.py
```diff
@@ -246,6 +246,34 @@ def test_unknown_os(self):
...
+        self.conf.inc_used_build()
+        assert self.conf.build_type == 'opt'
+
+    def test_fallback_routes(self):
+        routes = list(self.conf.tk_inbound_routes(
+            create_push('1a', TIMESTAMP_GECKO_V2)
+        ))
+        assert len(routes) == 3
+        assert routes[0].endswith('-opt')
+        assert routes[2].endswith('-fallback')
```
wlach: Could you assert on the specific return values here?

mozregression/fetch_configs.py
KwanEsq: Well if it does ever overflow and build_type is called we'll get `IndexError: list index out of range` anyway (which is what made me realise I need to add the mod operation to prevent it), so I think we should be fine.

tests/unit/test_fetch_configs.py
KwanEsq: Done.

mozregression/fetch_configs.py
KwanEsq: I think just to save the rebasing I'll just do a final commit that rips out all the buildbot stuff rather than doing it before.

mozregression/fetch_configs.py
wlach: Hmm, does this mean we might have an infinite loop? Seems like it would be better to bail out, wouldn't it?

KwanEsq: Well nothing should be iterating using the index anyway (since I realised I could just make `tk_inbound_routes` a generator and track the used build internally), they should just iterate `tk_inbound_routes` to get them all, and `tk_inbound_routes` doesn't actually use it, just iterates the build_types and set this.  I suppose we could have it just stop incrementing once it reaches the end.  I think the original reason I had it loop was so it would end-up "reset", and report it as looking for the original requested build rather than fallbacks on total failure, but I guess build_type isn't actually used in that instance, and it is reporting the last fallback route as the failure anyway.

wlach: Ok, sounds reasonable! I feel like we could simplify this area of the code a bunch, but that might require further refactoring. For now this looks like a good step forward.

KwanEsq:
> Please go ahead and merge this whenever you're ready!

Thanks a lot for the review!
I made two small changes in the final version:
- prefixed `inc_used_build` with a `_` to indicate it shouldn't be called outside of a config
- made `config.build_types` a tuple instead of a list, since it doesn't need to be appended to after setting (and reworked `set_build_type` appropriately).

PR #522, opened 2019-04-04
Bug 1541421 - Disable download of Fennec updates

staktrace:


wlach:
Double checking: did you test this?

staktrace:
yup. I ran `mozregression -n fennec --persist ~/Downloads/mozregression/ --launch 2018-04-01` and verified both that the fennec settings said 'disabled' for the download updates option and that after waiting a while it didn't actually attempt downloading. I didn't check what effect if any this has on desktop but it doesn't look like the pref is used on desktop so i assume no effect.

wlach:
Yup, I would not expect this to have an effect on desktop. Good enough for me! Merging.

PR #524, opened 2019-04-11
Improve support for version numbers on release repos

KwanEsq:
This should help with a number of ergonomic defects around the use of version numbers as build identifiers, and fixes a few different bugs: [bug 1308031](https://bugzilla.mozilla.org/show_bug.cgi?id=1308031), [bug 1494051](https://bugzilla.mozilla.org/show_bug.cgi?id=1494051), and [bug 1425164](https://bugzilla.mozilla.org/show_bug.cgi?id=1425164).

### DIFF
--- a/mozregression/releases.py
+++ b/mozregression/releases.py
@@ -125,6 +125,8 @@ def tag_of_beta(release):
     """
     if re.match(r'^\d+\.0b\d+$', release):
         return 'FIREFOX_%s_RELEASE' % release.replace('.', '_')
+    elif re.match(r'^\d+(\.0)?$', release):
+        return 'FIREFOX_RELEASE_%s_BASE' % release.replace('.0', '')
     else:
         raise UnavailableRelease(release)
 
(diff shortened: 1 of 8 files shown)


KwanEsq:
Switched from FIREFOX_BETA_#_END to FIREFOX_RELEASE_#_BASE for plain versions on beta.

mozregression/releases.py
```diff
@@ -106,6 +106,31 @@ def date_of_release(release):
...
+    """
+    Provide the mercurial tag of a beta release, suitable for use in place of a
+    hash
+    """
+    if re.match(r'^\d+\.0b\d+$', release):
+        return 'FIREFOX_%s_RELEASE' % release.replace('.', '_')
+    elif re.match(r'^\d+(\.0)?$', release):
+        return 'FIREFOX_RELEASE_%s_BASE' % release.replace('.0', '')
+    else:
+        raise UnavailableRelease(release)
```
wlach: Just as a point of interest, [buildhub](https://mozilla-services.github.io/buildhub/) would be an alternative option for getting this data, but I think hg tags are a better approach (probably less likely to go down).

KwanEsq: Ah, thanks for that, I might find it useful elsewhere.

KwanEsq:
Thanks!

PR #525, opened 2019-04-12
Correct error in aarch64 code in available_build_types

KwanEsq:
The previous code was incorrectly checking build_types for a `win64-aarch6464`
platform, which obviously doesn't exist.  Fortunately it doesn't matter at the
moment since win64-aarch64 isn't specified for any of the limited build types.

While this could have been fixed by just trimming the 64 off the end of the
substituted `os` string value (and switching the os check for the bits check),
switching to an explicit suffix should hopefully make it clearer what is
happening.


KwanEsq:
Will, given this is just fixing a dumb error in my previous code a review seemed like a poor use of your time, but if you do look over this and see anything you want me to change just let me know.

PR #526, opened 2019-04-14
Add comm-central to branches

oeekker:
Sorry, but I couldn't build this myself on my windows machine, but I believe this single line change is enough to add comm-central repository to the list of branches. It is also maybe time to remove mozilla-aurora and comm-aurora from the release branches, because they are out of use since May 2017.

### DIFF
--- a/mozregression/branches.py
+++ b/mozregression/branches.py
@@ -56,6 +56,7 @@ def create_branches():
     branches = Branches()
 
     branches.set_branch("mozilla-central", "mozilla-central")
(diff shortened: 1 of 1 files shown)


KwanEsq:
That's correct, thanks for doing this.  And aurora repos are indeed being removed in #524.
I should hopefully have some more improvements for using mozregression with Thunderbird in the future.

KwanEsq:
And just for reference, this fixes [bug 1510263](https://bugzilla.mozilla.org/show_bug.cgi?id=1510263).

PR #528, opened 2019-04-15
Bug 1422606 - Add support for opening app to a specific URL from GUI

KwanEsq:
Useful when needing to run a test with a specific site

I punted on generic arg in GUI support, since just a URL should cover most use cases.


KwanEsq:
Whoops, forgot to run `./check.py` with `-G`

wlach:
thanks for this! sorry for the lateness with merging

PR #532, opened 2019-04-30
Update dependencies

wlach:


wlach:
This paves the way for python 3, since all deps are now python 3 compatible.

PR #542, opened 2019-07-18
Bug 1567259 - Bail out of check_expand if not enough valid builds.

KwanEsq:
BuildRange's check_expand bails if it doesn't have enough builds to start with,
but it then filters out any invalid builds, which can reduce the amount below
the required threshold (2), and doesn't then check again. At that point
continuing will throw IndexErrors since it assumes there will be items at
indexes 0 and -1.

### DIFF
--- a/mozregression/build_range.py
+++ b/mozregression/build_range.py
@@ -179,6 +179,10 @@ class BuildRange(object):
         self._fetch((0, -1))
         self.filter_invalid_builds()
 
+        if len(self) < 2:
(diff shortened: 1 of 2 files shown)


wlach:
Awesome, would it be possible to write a unit test for this?

KwanEsq:
After a bit of guessing in the tests it seems like!  Manage to write one that produces the IndexError before my patch and passes afterwards :)

wlach:
Thank you!

PR #534, opened 2019-05-10
Bug 1464127: Make FirefoxLauncher always add --allow-downgrade to cmdargs

dblohm7:

PR #549, opened 2019-08-04
Revert "Bug 1464127: Make FirefoxLauncher always add --allow-downgrad…

KwanEsq:
…e to cmdargs (#534)"

For breaking URL arguments on older builds, and not fixing what it was trying
to.

This reverts commit 5b237ea1315df3d27891e71a5338d4cfb31851c1.

### DIFF
--- a/mozregression/launchers.py
+++ b/mozregression/launchers.py
@@ -331,13 +331,6 @@ class FirefoxLauncher(MozRunnerLauncher):
         super(FirefoxLauncher, self)._install(dest)
         self._disableUpdateByPolicy()
 
-    def _start(
-        self, profile=None, addons=(), cmdargs=(), preferences=None, adb_profile_dir=None,
-    ):
-        super(FirefoxLauncher, self)._start(
-            profile, addons, ["--allow-downgrade"] + cmdargs, preferences, adb_profile_dir,
-        )
-
 
 class ThunderbirdRegressionProfile(ThunderbirdProfile):
     """


wlach:
@KwanEsq am I correct in thinking that something else (mozrunner?) adds --allow-downgrade? I trust your judgement on this, feel free to merge.

KwanEsq:
Bah, sorry for not getting back to this.
So basically this was partially my fault for misdiagnosing [bug 1550806](https://bugzilla.mozilla.org/show_bug.cgi?id=1550806), and not realising the `--allow-downgrade` stuff had already landed in #510.  In the cases where it's needed it is indeed already set, by main.py https://github.com/mozilla/mozregression/blob/14d0d7388b8123bcfe0f4cf64ecfc72cf0466b25/gui/mozregui/build_runner.py#L150-L151
and gui's build_runner.py
https://github.com/mozilla/mozregression/blob/14d0d7388b8123bcfe0f4cf64ecfc72cf0466b25/mozregression/main.py#L54-L63
It might not be that big an issue but for a while Firefox was incorrectly swallowing any arguments after `--allow-downgrade` as a value to it (see [bug 1527374](https://bugzilla.mozilla.org/show_bug.cgi?id=1527374)), so where it's set first here it can swallow the URL argument, which is obviously inconvenient when trying to test against one.

Rebased version coming shortly.

(Also apologies, but I can no longer merge myself. I suffered a mishap transferring my 2FA to a new phone, and as part of the recovery process 2FA was briefly removed, which rightfully automatically removed me as a collaborator since Mozilla require 2FA)

wlach:
This makes sense, thanks!

PR #550, opened 2019-09-11
Bug 1404520 - Raise a helpful error when missing required build info

mars-f:
The NightlyInfoFetcher._fetch_build_info_from_url() method constructs
a dict with at least one key, "build_url", that is passed back
to the caller. However it was possible under specific circumstances to
build a dict where the "build_url" key was missing, causing later parts
of the program to fail.

This code adds a check for the "build_url" key and raises a helpful
error if the key is missing.


wlach:
@mars-f thank you for these! would you mind quickly rebasing these PR's on current master and repushing? there was an error that was preventing us from running ci

PR #551, opened 2019-09-11
Bug 1404520 - Fix the Linux ASAN build file regex

mars-f:
Fix the regex used to find Linux ASAN build files.


wlach:
I am not able to verify this, but will take your word for it. :)

PR #558, opened 2019-11-13
Bug 1589881 - Don't fail if a pushlog URL 404's

mhowell-mozilla:
Currently if a push doesn't exist in a repo we look for it in, the 404 error results in a `MozRegressionError` exception, which is fatal. That means we're failing out if a push doesn't exist in any one of the repos we search, even if it does exist in another (for example, an autoland push that's been merged to central, but not yet back to inbound). So, change that exception to an `EmptyPushlogError`, which isn't fatal.

I'd expect this to be a pretty common problem now that merges to mozilla-inbound have been slowed down and commit access to it is being sharply limited.

### DIFF
--- a/mozregression/json_pushes.py
+++ b/mozregression/json_pushes.py
@@ -3,7 +3,7 @@ import datetime
 
 from mozlog import get_proxy_logger
 
-from mozregression.errors import MozRegressionError, EmptyPushlogError
+from mozregression.errors import EmptyPushlogError
 from mozregression.network import retry_get
 from mozregression import branches
 from mozregression.dates import is_date_or_datetime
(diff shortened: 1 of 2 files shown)


mozregression/json_pushes.py
```diff
@@ -69,9 +69,9 @@ def pushes(self, **kwargs):
 
         response = retry_get(url)
         if response.status_code == 404:
-            raise MozRegressionError(
-                "The url %r returned a 404 error. Please check the"
-                " validity of the url." % url
+            raise EmptyPushlogError(
+                "The url %r returned a 404 error. The push may not be in"
+                " this repo (e.g., not merged yet)." % url
```
wlach: 👍 I think JsonPushes is the place to handle this. It looks like if we open this URL:

https://hg.mozilla.org/integration/mozilla-inbound/json-pushes?changeset=2f19e7b646e0a52fa855b75c868f0a3f3a990ad3&full=1

We get this raw json response:

```json
{
"error": "unknown revision '2f19e7b646e0a52fa855b75c868f0a3f3a990ad3'"
}
```

Do you think you could check the response data for a string of this kind before returning EmptyPushLogError? That way we'll fail correctly in case there is a 404 for some other reason (e.g. server misconfiguration)

mhowell-mozilla: I'll do that. I think I still end up not needing MozRegressionError because raise_for_status() will do effectively the same thing if we 404 for another reason?

wlach: Ah yes, you are right. You can remove the import from this file, and the linter should start passing again

wlach:
Looks good except it looks like your editor added dos line endings to the file. Attempting to fix.

mhowell-mozilla:
Oh, interesting, thanks for fixing that. Sublime seems to think that all the files already had CRLF, so Git must be misconfigured. And I'm on Windows so I wouldn't have noticed. ;)

PR #563, opened 2020-01-20
Bug 1606902 - Set default integration branch to autoland

wlach:
In some cases (e.g. bisecting GeckoView apps) we would attempt
to find builds on "mozilla-inbound" by default, which is now
discontinued.


wlach:
I had a bit of difficulty testing Android apps on my machine, but it looks like the error botond noticed in the original bug is fixed at least (it downloads the build fine)

PR #565, opened 2020-01-24
Bug 1437591 - Add --arch argument to use specific CPU package.

makotokato:
Actually, `mozregression` only supports arm package for Fennec and GVE. But I would like to support other architecture package such as x86_64 since it works on emulator.


wlach:
Let's close this in favour of #644, thanks for submitting the PR though!

PR #569, opened 2020-03-16
Bug 1581647 - Add glean-based telemetry to mozregression

wlach:


wlach:
@mdboom could you take a look at this? I think we're about ready to go here aside from data review and the need for a 32-bit Windows wheel

metrics.yaml
```diff
@@ -0,0 +1,29 @@
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
+
+usage:
```
Dexterp37: Fly-by: glean parser can be used to generate the _required_ metric documentation. In other platforms/bindings, this is usually deep integrated within the build system (e.g. Gradle on Android), making it transparent to the user and guaranteeing that it's almost always up to date. While I'm not sure we already have a good story about that on Python, there's still value in manually generating the docs.

In this specific case it might be worthwhile to add a comment at the top of the file explaining how to generate docs.

pings.yaml
```diff
@@ -0,0 +1,13 @@
+---
+$schema: moz://mozilla.org/schemas/glean/pings/1-0-0
```
Dexterp37: The same documentation comment applies to this file as well.

metrics.yaml
Dexterp37: I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1624559) for further discussions.

wlach: Thanks Alessio! I think this is a little less important for mozregression where we're only collecting a small number of things and it's easy to keep things updated by hand (P.S. some docs in #572 if you want to take a look) but I agree it would be nice to get this solved more generally.

Dexterp37: My pleasure! I would encourage you to use still generate and commit the generated docs (even for a small number of metrics), since the format is the same across all products and you get important details (who owns the metric? where is the data review?) along the description :)

gui/mozregui/build_runner.py
```diff
@@ -172,6 +173,11 @@ def start(self, fetch_config, options):
         self.thread.start()
         # this will be called in the worker thread.
         QTimer.singleShot(0, action)
+        # an action = instance of mozregression usage, so send
+        # a usage ping (if telemetry is enabled)
+        if get_prefs()['enable_telemetry']:
```
mdboom: No need to check this value here -- you can let Glean's enable / disable handle this.

mozregression/telemetry.py
```diff
@@ -0,0 +1,55 @@
...
+    mozregression_path = Path.home() / '.mozilla' / 'mozregression'
+    Glean.initialize(
+        application_id="mozregression",
+        application_version=__version__,
+        upload_enabled=True,
+        configuration=Configuration(
+            ping_tag="mozregression-test-tag"
+        ),
+        data_dir=mozregression_path / "data"
+    )
```
mdboom: I would separate out initialization from submitting the ping here.  `Glean.initialize` [should always be called on every startup](https://mozilla.github.io/glean/book/user/general-api.html#initializing-the-glean-sdk) -- whether telemetry is enabled or not -- so that it can take care of sending any pending pings, and the deletion ping can be handled if necessary.

mozregression/telemetry.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # if we got this far, we can assume telemetry is enabled
+    Glean.set_upload_enabled(True)
```
mdboom: This is redundant to the `upload_enabled` argument to `Glean.initialized`.  However, this should be called from the callback to the telemetry checkbox in the GUI so that Glean can clear metrics, reset the client id and send a deletion ping request if necessary.

mozregression/telemetry.py
```diff
@@ -0,0 +1,55 @@
...
+
+    pings = load_pings("pings.yaml")
+    metrics = load_metrics("metrics.yaml")
```
mdboom: Since you might send multiple pings per run, you might as well do this parsing at startup and save it in a global variable rather than parsing every time.

Also, you probably want to install these files as package data, and then retrieve them at runtime.  I think this will fail if running `mozregression` from a current working directory that doesn't contain these files.  See [including package data](https://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files).  There's other ways to do that in Python if you don't want to use setuptools -- my point is just that they should be installed alongside the code so you don't have to run from a particular directory at run time.

mozregression/telemetry.py
```diff
@@ -0,0 +1,55 @@
...
+        # sending telemetry pings is python3-only, since
+        # it relies on Glean
+        LOG.info("Not using python 3+, not sending telemetry ping")
+        return
+
+    LOG.info("Sending anonymized telemetry recording mozregression "
+             "usage, learn more at: https://mozilla.github.io/mozregression/telemetry.html")
+
+    p = Process(target=_send_telemetry_ping, args=(variant, appname))
+    p.start()
```
mdboom: This totally makes sense to not block on I/O.  This might change as part of [a bug to provide built-in multiprocessing for Glean](https://bugzilla.mozilla.org/show_bug.cgi?id=1604191), but we can come back around to this then.

mozregression/telemetry.py
wlach: Hmm, so Glean submission is handled as a *single* process launched from the main one in which mozregression is running with multiprocessing. In this model, there's no other way than to do initialization + sending in the same method. I think there's a few advantages to this approach:

1. There is no chance that submitting the ping will interfere with mozregression's own operations.
2. If used as a library, the calling process can do its own telemetry -- this is especially a concern with the mach wrapper, which calls into mozregression this way and will probably be doing its own telemetry: https://searchfox.org/mozilla-central/rev/202a285024f174c2d2bf2152d9cba90a03723eab/tools/mach_commands.py#398

Passing a "glean enabled" flag to the process is easy, of course. If I do that, will it operate correctly? I realize this goes against the grain of how Glean was envisioned to work but I'd rather prioritize making sure mozregression operates with no issue for its user.

mdboom: I think for the multiprocessing issue -- which basically boils down to not having Glean block the application -- we probably need to handle that at the Glean level, if not doing it "all in one go" like this.  I think it's possible -- we do this on other platforms, just that Python is extra annoying because with the GIL threading isn't likely to work well and we have to use multiprocessing.  I have some ideas for this already, and this probably pushes them up in priority.  This would negate the need for passing a "glean enabled" flag to the process -- we should be able to handle that kind of thing within the Glean SDK.

The case of mozregression being both a library and an application depending on the context is new, but should fundamentally not be a problem as long as *only* the application calls Glean initialize, given the current design.  When using mozregression as a library, it can still set metrics and send pings even if the library in which it embeds it is the one calling `Glean.initialize`.

So I think the action item is -- I'll start looking at adding built-in multiprocessing to Glean.  If that looks feasible, we could refactor this to separate out init and recording/submitting as discussed.  If that turns into a quagmire, this approach is probably a good workaround for now.

wlach: In the case of mach won't the application id be different/wrong in the case of the call to `Glean.initialize`? I assume mach will be passing in something like `mach` for this parameter.

mdboom: Indeed that's true.  Up until now, we've considered data as "sent by applications", even if it's from a library embedded in an application.  For example, application-services sync data ends up in tables related to Fenix.  I think in this case it's less useful to think of mozregression data sent from within mach as different from data sent from outside of mach.  This probably will require a little design discussion to get right -- I think we could still have a single Glean instance per application / process, but maybe we allow for pings that are related to a library is directly and not an application id.

Thanks for poking holes in all of our design assumptions ;)

Cc @dexterp37

wlach: Ok, I'm going to leave this aspect as-is for now.

mozregression/telemetry.py
wlach: Sending the deletion ping doesn't seem to be working (at least with test tags and the debug ping viewer), but I've deleted this line.

mozregression/telemetry.py
Dexterp37: > Thanks for poking holes in all of our design assumptions ;)

Indeed! Thanks @wlach ! (for future reference, @mdboom filed a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1624695) about this).

metrics.yaml
Dexterp37: @wlach I think this comment slipped through. I would really strongly encourage you to generate docs with the glean parser. These are the doc-o-types that were verified by other teams to make sure they are consistent, readable and have all the information require to make them useful.

You can do so by manually running:

`glean_parser translate -f markdown -o docs/ metrics.yaml pings.yaml`

wlach: Done! Thanks for the clarification.

docs/documentation/telemetry.md
```diff
@@ -0,0 +1,31 @@
...
+title: Telemetry
+doc_order: 100
+description: About the use of Telemetry in mozregression
+---
+
+# Telemetry
+
+Unless explicitly disabled, mozregression will send an anonymized telemetry ping (a small JSON-document) to Mozilla's servers on each use (command-line invocation in the case of console mozregression, each bisection or single run in the case of the GUI).
+
+This data is anonymized and not personally identifiable, it contains only:
```
Dexterp37: I would be very cautious in stating this: while this ping doesn't contain PII, there's no active effort into performing deep anonymization.

docs/documentation/telemetry.md
```diff
@@ -0,0 +1,31 @@
...
+
+- A random client id (e.g. `5a26b9f7-c025-4858-abce-63ca73ceebf0`)
+- Some very basic information on the environment in which mozregression
```
Dexterp37:
```suggestion
- [Minimal information](https://mozilla.github.io/glean/book/user/pings/index.html#ping-sections) on the environment in which mozregression
```

Dexterp37: I think glean_parser should do a better job at documenting that pings also contain some information about the device. I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1625412) to improve our docs so that you don't have to mention it.

docs/documentation/telemetry.md
```diff
@@ -0,0 +1,31 @@
...
+  is running (e.g. operating system)
+- Some other metrics designed to measure and understand the nature of
+  usage of mozregression (for example, which applications are being bisected) -- for an up-to-date list, please see this [automatically
+  generated report](https://github.com/mozilla/mozregression/blob/master/docs/glean/metrics.md).
+
+The purpose of this data gathering is only to improve mozregression itself by understanding the scope of its environment and usage, and
```
Dexterp37: This sould really go into the `pings.yaml` ping description, so that it shows up in the autogenerated docs.

docs/documentation/telemetry.md
```diff
@@ -0,0 +1,31 @@
...
+description: About the use of Telemetry in mozregression
+---
+
+# Telemetry
+
+Unless explicitly disabled, mozregression will send an anonymized telemetry ping (a small JSON-document) to Mozilla's servers on each use (command-line invocation in the case of console mozregression, each bisection or single run in the case of the GUI).
+
+This data is anonymized and not personally identifiable, it contains only:
+
+- A random client id (e.g. `5a26b9f7-c025-4858-abce-63ca73ceebf0`)
```
Dexterp37: If you link to the glean docs below, then you can also drop this sentence.

docs/documentation/telemetry.md
wlach: Good point, i'll remove the term "anonymization"

docs/documentation/telemetry.md
wlach: Done

docs/documentation/telemetry.md
wlach: Great, I'll just take this out and refer people to the autogenerated docs

wlach:
I refactored the GUI so it uses Glean in the standard way (i.e. no weird multiprocess forking) as I think that might lead to weird issues in a Windows environment (especially with pyinstaller as a variable).

I also dropped support for python 2.7 in mozregression itself (in a seperate PR), as I think that will simplify this.

With these changes, I think this is almost ready to go, just waiting for https://bugzilla.mozilla.org/show_bug.cgi?id=1624695 (so that we send deletion requests in the command line variants when telemetry is disabled between runs) and privacy review per https://bugzilla.mozilla.org/show_bug.cgi?id=1581647#c15

@mdboom @Dexterp37 want to take one last look?

Dexterp37:
> I also dropped support for python 2.7 in mozregression itself (in a seperate PR), as I think that will simplify this.

OT: Doesn't `mach` require support for that, still?

docs/documentation/telemetry.md
```diff
@@ -0,0 +1,27 @@
+---
+layout: documentation
+title: Telemetry
+doc_order: 100
+description: About the use of Telemetry in mozregression
+---
+
+# Telemetry
+
+Unless explicitly disabled, mozregression will send an anonymized telemetry "ping" (a small JSON-document) to Mozilla's servers on each use (command-line invocation in the case of console mozregression, each bisection or single run in the case of the GUI).
```
Dexterp37: The anonymization part is still around :-P See [this comment](https://github.com/mozilla/mozregression/pull/569#discussion_r399119612).

Dexterp37:
```suggestion
Unless explicitly disabled, mozregression will send a telemetry ["ping"](https://mozilla.github.io/glean/book/user/pings/index.html) to Mozilla's servers on each use (command-line invocation in the case of console mozregression, each bisection or single run in the case of the GUI).
```

mozregression/metrics.yaml
```diff
@@ -0,0 +1,29 @@
...
+  variant:
+    type: string
+    description: >
+      The variant of mozregression used to perform the bisection (gui,
+      console, mach, etc.)
+    notification_emails: [wlachance@mozilla.com]
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1581647
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1581647
```
Dexterp37: Please link to the specific comment with the **data-review response**.

mozregression/metrics.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - usage
+  app:
+    type: string
+    description: >
+      The name of the app being used (firefox, gve, etc.)
+    notification_emails: [wlachance@mozilla.com]
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1581647
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1581647
```
Dexterp37: Please link to the specific comment with the **data-review response**.

mozregression/pings.yaml
```diff
@@ -0,0 +1,18 @@
...
+    shared except in aggregated form. Although not a consumer product,
+    mozregression strives to follow Mozilla's general guidelines on
+    [lean data practices](https://www.mozilla.org/en-US/about/policy/lean-data/).
+  include_client_id: true
+  notification_emails:
+    - wlachance@mozilla.com
+  bugs:
+    - https://bugzilla.mozilla.org/show_bug.cgi?id=1581647
+  data_reviews:
+    - https://bugzilla.mozilla.org/show_bug.cgi?id=1581647
```
Dexterp37: Please link to the specific comment with the **data-review response**.

docs/documentation/telemetry.md
```diff
@@ -0,0 +1,27 @@
...
+
+The purpose of this data gathering is only to improve mozregression itself by understanding the scope of its environment and usage, and
+will not be broadly shared except in aggregated form. Although not a consumer product, mozregression strives to follow Mozilla's general guidelines on [lean data practices](https://www.mozilla.org/en-US/about/policy/lean-data/).
+
+## How to disable
+
+We encourage you to leave Telemetry enabled -- if we know that people
+are getting value out of mozregression, it provides an incentive to
+make it better! That said, disabling mozregression is easy. In the
+command-line variant, you can set `enable-telemetry` to no inside your [configuration file](./configuration.md). In the GUI version, simply untick the box "Enable Telemetry" in the preferences dialog (after clicking "Show Advanced Options").
```
Dexterp37:
```suggestion
command-line variant, you can set `enable-telemetry` to `no` inside your [configuration file](./configuration.md). In the GUI version, simply untick the box "Enable Telemetry" in the preferences dialog (after clicking "Show Advanced Options").
```

wlach:
> > I also dropped support for python 2.7 in mozregression itself (in a seperate PR), as I think that will simplify this.
> 
> OT: Doesn't `mach` require support for that, still?

Nope! mach supports running commands in a python3 virtualenv now, and I made mozregression do that in [bug 1616584](https://bugzilla.mozilla.org/show_bug.cgi?id=1616584)

PR #579, opened 2020-03-31
virtualenv for appveyor

wlach:


wlach:
This actually only saves a few k, the main optimization I wanted to make (1) didn't work when run (2) didn't even save much space (it looks like QtNetwork was the only unnecessary(?) dependency that pyinstaller was including). May as well merge I guess...

PR #639, opened 2020-05-14
Bug 1631163 - Improve version compare

jay:
Prior to this change the version compare did not accurately compare
development versions against release versions, and as a result dev users
may have been shown an erroneous upgrade notice in the console or gui.

For example, dev version 4.0.5.dev8+gedaf3e9.d20200514 comes after
release version 4.0.4, and now there is no notice to upgrade to the
latter.

Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1631163

Closes #xxx

---

Only thing I'm unsure about is I think packaging comes standard

/cc @wlach

### DIFF
--- a/gui/mozregui/check_release.py
+++ b/gui/mozregui/check_release.py
@@ -1,3 +1,5 @@
+from packaging import version
+
 from PySide2.QtCore import QObject, Qt, QThread, QUrl, Slot
 from PySide2.QtGui import QDesktopServices
 from PySide2.QtWidgets import QLabel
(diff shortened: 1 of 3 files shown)


jay:
Sure I can take a look this week. I think I had some problem running testing due to missing mock or something.

jay:
> I think I had some problem running testing due to missing mock or something.

I haven't solved the mock problem. I [posted an update](https://bugzilla.mozilla.org/show_bug.cgi?id=1536324#c10) in the bug thread. I will make an attempt to fix the CI errors, they look legit, but I may not get it right the first time since I don't have the ability to run the tests.

PR #641, opened 2020-05-14
Bug 1638073 - Support android shippable index's

Callek:
I haven't formally tested this yet, but will be needed once the stack on https://bugzilla.mozilla.org/show_bug.cgi?id=1614970 lands, I'll comment here when that happens.

### DIFF
--- a/mozregression/fetch_configs.py
+++ b/mozregression/fetch_configs.py
@@ -412,8 +412,12 @@ class FennecIntegrationConfigMixin(IntegrationConfigMixin):
             elif push.timestamp >= TIMESTAMP_FENNEC_API_15:
                 tk_name = "android-api-15"
         for build_type in self.build_types:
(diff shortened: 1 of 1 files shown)


wlach:
@Callek I don't know much about the planned taskcluster route updates -- will this change break finding historical shippable builds? Or should the new route provide those automatically?

Callek:
> @Callek I don't know much about the planned taskcluster route updates -- will this change break finding historical shippable builds? Or should the new route provide those automatically?

This shouldn't break much about historical shippable builds, the full changeset I have (give or take bitrot updates) is at https://phabricator.services.mozilla.com/D74510

The main change for android is nightly and pgo are going away and we're adding shippable. android tasks are also primarily geckoview in central right now - and mozregression doesn't even seem to track pgo or nightly right now for android, so I think this ends up being only additive.

Callek:
@wlach we can merge this now, thanks!

PR #644, opened 2020-05-15
Bug 1437591 - Add support for x86_64 GVE builds

staktrace:
I just cobbled this together quickly as a minimum viable thing for what I needed. Happy to clean it up as you see fit. I only saw the PR in #565 after I wrote and was kind of surprised at how similar they were, but that one seems more complicated. Also the problem with using `mozinfo.processor` as the default is that it seems to be based on the host machine which we don't really care about so much for Android. My PR keeps the `arch` separate from `processor`.

### DIFF
--- a/mozregression/cli.py
+++ b/mozregression/cli.py
@@ -267,6 +267,13 @@ def create_parser(defaults):
     )
 
     parser.add_argument(
+        "--arch",
+        choices=("arm", "x86_64"),
+        default=None,
+        help=("Force x86_64 build (only applies to GVE app). Default: arm"),
+    )
+
+    parser.add_argument(
         "-c",
         "--command",
         help=(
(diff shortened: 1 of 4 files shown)


mozregression/cli.py
```diff
@@ -550,7 +557,15 @@ def validate(self):
 
         user_defined_bits = options.bits is not None
         options.bits = parse_bits(options.bits or mozinfo.bits)
-        fetch_config = create_config(options.app, mozinfo.os, options.bits, mozinfo.processor)
+        if options.arch is not None:
+            self.logger.info("app is %s" % options.app)
+            if options.app != "gve":
+                self.logger.info("arch option being ignored for non-GVE app.")
```
wlach, line 563: Let's make this a warning

```suggestion
                self.logger.warning("--arch ignored for non-GVE app.")
```

mozregression/cli.py
```diff
@@ -550,7 +557,15 @@ def validate(self):
 
         user_defined_bits = options.bits is not None
         options.bits = parse_bits(options.bits or mozinfo.bits)
-        fetch_config = create_config(options.app, mozinfo.os, options.bits, mozinfo.processor)
+        if options.arch is not None:
+            self.logger.info("app is %s" % options.app)
```
wlach, line 561: This looks like a debugging statement that snuck in

```suggestion
```

wlach:
It would be nice to expose this in the GUI as well -- many mozregression users are using that, not asking you to do that here but perhaps we can file a bug to followup

staktrace:
Patch updated to address nits. I also added a second patch which updates the GUI. Tested by manually running it and exercising a bunch of the codepaths, not sure if there's a better way to do that.

Ready for re-review.

PR #647, opened 2020-05-18
Build GUI using python 3.8

wlach:


wlach:
This seems to work pretty well, let's go ahead and merge it.

PR #651, opened 2020-05-20
Remove some unnecessary octals modernize introduced

KwanEsq:
These were introduced in f6f16683ebb8 since the month and date numbers were
zero-padded for alignment in Python 2, but it's invalid syntax in 3.


wlach:
🤦 thanks!

PR #669, opened 2020-06-15
Bug 1401649 - Add support for bisect l10n nightly builds

KwanEsq:
This is a bit hacky, but gets the job done, and I think a less-hacky way would involve a greater re-factoring than I am willing or able to attempt.

### DIFF
--- a/gui/mozregui/build_runner.py
+++ b/gui/mozregui/build_runner.py
@@ -1,7 +1,7 @@
 from PySide2.QtCore import QObject, QThread, QTimer, Signal, Slot
 
 from mozregression.download_manager import BuildDownloadManager
-from mozregression.errors import LauncherError
+from mozregression.errors import LauncherError, MozRegressionError
 from mozregression.network import get_http_session
 from mozregression.persist_limit import PersistLimit
 from mozregression.telemetry import send_telemetry_ping
@@ -153,6 +153,13 @@ class AbstractBuildRunner(QObject):
         # Thunderbird will fail to start if passed an URL arg
         if options.get("url") and fetch_config.app_name != "thunderbird":
             launcher_kwargs["cmdargs"] += [options["url"]]
+
+        # Lang only works for firefox-l10n
+        if options.get("lang"):
+            if options["application"] == "firefox-l10n":
+                fetch_config.set_lang(options["lang"])
+            else:
+                raise MozRegressionError("Invalid lang argument")
 
         self.worker = self.worker_class(fetch_config, self.test_runner, self.download_manager)
         # Move self.bisector in the thread. This will
--- a/gui/mozregui/ui/intro.ui
+++ b/gui/mozregui/ui/intro.ui
@@ -59,7 +59,7 @@
     </widget>
    </item>
    <item row="4" column="0">
-    <widget class="QLabel" name="label_3">
+    <widget class="QLabel" name="repository_label">
      <property name="toolTip">
       <string>you can choose a specific repository name here</string>
      </property>
@@ -79,6 +79,26 @@
     </widget>
    </item>
    <item row="5" column="0">
+    <widget class="QLabel" name="lang_label">
+      <property name="toolTip">
+        <string>enter the code of the lang you want a build for</string>
+      </property>
+      <property name="text">
+        <string>Lang</string>
+      </property>
+    </widget>
+   </item>
+   <item row="5" column="1">
+     <widget class="QLineEdit" name="lang">
+       <property name="toolTip">
+        <string>Lang code such as he, ar, zh-TW</string>
+      </property>
+      <property name="placeholderText">
+        <string>The lang code (e.g. de)</string>
+      </property>
+     </widget>
+   </item>
+   <item row="6" column="0">
     <widget class="QLabel" name="url_label">
      <property name="toolTip">
       <string>enter a URL to pass to the app</string>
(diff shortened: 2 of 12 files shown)


wlach:
@KwanEsq this looks like a promising start, however I'm not seeing the lang option in the dialog in the GUI? I double checked that changes are actually being picked up by modifying one string.

Screen Shot 2020-06-17 at 10 32 16 AM

KwanEsq:
So since it's a whole other fetch config you have to select Application firefox-l10n to get offered the Lang option.  I had wondered about auto-switching app if lang was provided a value on the CLI, though hadn't thought much about it in the GUI, but I suppose it's possible (though wouldn't allow hiding repository like it does now).

Another thing I had pondered, and @ItielMaN raised as feedback after using it, was providing a list of langs to select from rather than a free-form text entry. I think it should be doable with https://product-details.mozilla.org/1.0/languages.json I mainly hadn't bothered at first out of probably excessive paranoia at the possibility of that preventing someone bisecting on a lang that had builds but was somehow missing from the constructed list.

(As an aside the macOS dialog looks weird with those super-narrow inputs, I guess because of [`QFormLayout::FieldsStayAtSizeHint`](https://doc.qt.io/qt-5/qformlayout.html#FieldGrowthPolicy-enum): <q cite="https://doc.qt.io/qt-5/qformlayout.html#FieldGrowthPolicy-enum">The fields never grow beyond their effective size hint. This is the default for QMacStyle.</q>)

wlach:
> So since it's a whole other fetch config you have to select Application firefox-l10n to get offered the Lang option. I had wondered about auto-switching app if lang was provided a value on the CLI, though hadn't thought much about it in the GUI, but I suppose it's possible (though wouldn't allow hiding repository like it does now).

I see. I think this is reasonable-- at least we can tell people how to do this in the case that they want to.

> Another thing I had pondered, and @ItielMaN raised as feedback after using it, was providing a list of langs to select from rather than a free-form text entry. I think it should be doable with https://product-details.mozilla.org/1.0/languages.json I mainly hadn't bothered at first out of probably excessive paranoia at the possibility of that preventing someone bisecting on a lang that had builds but was somehow missing from the constructed list.

This would be a nice addition.

> (As an aside the macOS dialog looks weird with those super-narrow inputs, I guess because of [`QFormLayout::FieldsStayAtSizeHint`](https://doc.qt.io/qt-5/qformlayout.html#FieldGrowthPolicy-enum): "The fields never grow beyond their effective size hint. This is the default for QMacStyle.")

Aha! Thank you, that was driving me crazy. Changing this field in Qt designer solves this:

image

I can do up a PR for this at some point, if you don't get to it before me. :)

gui/mozregui/ui/intro.ui
```diff
@@ -79,6 +79,26 @@
...
+      </property>
+    </widget>
+   </item>
+   <item row="5" column="1">
+     <widget class="QLineEdit" name="lang">
+       <property name="toolTip">
+        <string>Lang code such as he, ar, zh-TW</string>
+      </property>
+      <property name="placeholderText">
+        <string>The lang code</string>
```
wlach, line 97:
```suggestion
        <string>The lang code (e.x. de)</string>
```

gui/mozregui/build_runner.py
```diff
@@ -154,6 +154,10 @@ def init_worker(self, fetch_config, options):
         if options.get("url") and fetch_config.app_name != "thunderbird":
             launcher_kwargs["cmdargs"] += [options["url"]]
 
+        # Lang only works for firefox-l10n
+        if options.get("lang") and options["application"] == "firefox-l10n":
```
wlach, line 158: We should probably error out if lang is specified but the application is not `firefox-l10n`

gui/mozregui/wizard.py
```diff
@@ -147,6 +148,18 @@ def _set_fetch_config(self, index):
...
             self.ui.url_label.show()
 
+        # lang only makes sense for firefox-l10n, and repo doesn't
+        if app_name == "firefox-l10n":
+            self.ui.lang.show()
+            self.ui.lang_label.show()
+            self.ui.repository.hide()
+            self.ui.repository_label.hide()
+        else:
+            self.ui.lang.hide()
```
wlach, line 158: I wonder if we would be better off making it insensitive/grey rather than hiding it? I think it would be a little less confusing than have UI elements appear/disappear (I believe this is the recommended practice according to many UI style guides)

mozregression/cli.py
```diff
@@ -248,6 +248,12 @@ def create_parser(defaults):
         help="application name. Default: %(default)s.",
     )
 
+    parser.add_argument(
+        "--lang",
+        metavar="[ar|es-ES|he|ja|zh-CN|...]",
+        help=("build language. Only valid when app is " "firefox-l10n."),
```
wlach, line 254:
```suggestion
        help=("build language. Only valid when app is firefox-l10n."),
```

gui/mozregui/ui/intro.ui
KwanEsq: Thanks, done (I think I thought about when writing it, but decided to match Repository by only having it in the title).

gui/mozregui/build_runner.py
KwanEsq: Good idea, done.

gui/mozregui/wizard.py
KwanEsq: Ah, yes I like this better.  I wasn't happy with the disappearing myself (especially since the hidden ones seem to leave their margins around, or something else that affects the vertical gaps between options).

Should do a follow up to do the same with the rest of the options (arch, URL etc.)

mozregression/cli.py
KwanEsq: Ah, looks like an artefact of rebasing + auto-formatting, good catch thanks.

PR #684, opened 2020-07-09
Bug 1486004 - Support unrooted Android devices using mozdevice 4.0.0

bclary:
I tested this change locally with an unrooted Pixel2 and successfully completed a mozregression bisection. It has a number of dependency changes which have not be committed upstream on mozilla-central and the new packages are not yet available, but I wanted to get this PR out there so we can plan on the next release of mozregression which will support unrooted devices.


staktrace:
I tested this PR on a rooted pixel2 and an unrooted Sony Xperia XZ1C and it worked as intended on both. So this SGTM.

PR #685, opened 2020-07-09
Bug 1651760 - Save GVE architecture in persisted filename

staktrace:
This prevents architecture mismatches when installing a persisted APK.


staktrace:
@wlach thanks for the pointer, patch updated with tests.

PR #686, opened 2020-07-09
Bug 1651401 - Add additional metrics on what builds are being retrieved

wlach:
This will tell us the usage patterns of mozregression, which
helps inform what types of builds we need to keep around. Specifically,
track:

* build_type (e.g. shippable, opt, etc.)
* bisection/launch dates (if specified)

### DIFF
--- a/mozregression/metrics.yaml
+++ b/mozregression/metrics.yaml
@@ -30,12 +30,12 @@ usage:
   build_type:
     type: string
     description: >
-      The build type being bisected (opt, shippable, etc.)
+      The build type being bisected (asan, debug, opt, pgo, shippable, ...)
     notification_emails: [wlachance@mozilla.com]
     bugs:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1651401
     data_reviews:
-      - https://bugzilla.mozilla.org/show_bug.cgi?id=1651401
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1651401#c5
     expires: never
     send_in_pings:
       - usage
@@ -49,7 +49,7 @@ usage:
     bugs:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1651401
     data_reviews:
-      - https://bugzilla.mozilla.org/show_bug.cgi?id=1651401
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1651401#c5
     expires: never
     send_in_pings:
       - usage
@@ -63,7 +63,7 @@ usage:
     bugs:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1651401
     data_reviews:
-      - https://bugzilla.mozilla.org/show_bug.cgi?id=1651401
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1651401#c5
     expires: never
     send_in_pings:
       - usage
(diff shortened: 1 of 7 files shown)


mozregression/metrics.yaml
```diff
@@ -27,3 +27,57 @@ usage:
...
       - usage
+  build_type:
+    type: string
+    description: >
+      The build type being bisected (opt, shippable, etc.)
+    notification_emails: [wlachance@mozilla.com]
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1651401
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1651401
```
wlach, line 38: Note to self: update these when data review is complete

wlach:
@mdboom can you review the python parts? not sure who to ask
@Dexterp37 just thought I'd CC you on this since you've had useful feedback on Glean things in the past :)

mozregression/metrics.yaml
```diff
@@ -27,3 +27,57 @@ usage:
     expires: never
     send_in_pings:
       - usage
+  build_type:
+    type: string
+    description: >
+      The build type being bisected (opt, shippable, etc.)
```
mdboom, line 33: Is the set of values small?  If so, maybe list them all here...?  Mostly just thinking about the data scientist that might look at the data someday...  (We need an enumerated type in Glean, but we don't yet have one).

mozregression/telemetry.py
```diff
@@ -6,41 +7,56 @@
...
 
 def initialize_telemetry(upload_enabled, allow_multiprocessing=False):
     mozregression_path = Path.home() / ".mozilla" / "mozregression"
     Glean.initialize(
         application_id="org.mozilla.mozregression",
         application_version=__version__,
         upload_enabled=upload_enabled,
-        configuration=Configuration(allow_multiprocessing=allow_multiprocessing),
+        configuration=Configuration(
+            allow_multiprocessing=allow_multiprocessing, ping_tag="mozregression-test"
```
Dexterp37, line 28: Nit: don't forget to remove `ping_tag` before you ship. We're working to make this less weird by allowing configuring debug options using environment vars, see [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1646706).

mozregression/telemetry.py
```diff
@@ -6,41 +7,56 @@
...
+        ),
         data_dir=mozregression_path / "data",
     )
 
 
-def _send_telemetry_ping(variant, appname):
-    METRICS.usage.variant.set(variant)
-    METRICS.usage.app.set(appname)
+def _send_telemetry_ping(metrics):
+    print(metrics)
```
Dexterp37, line 35: nit: is it intentional to leave this here?

tests/unit/test_main.py
```diff
@@ -258,7 +259,18 @@ def create_app(fetch_config, options):
         except SystemExit as exc:
             self.assertEqual(send_telemetry_ping_oop.call_count, 1)
             self.assertEqual(
-                send_telemetry_ping_oop.call_args, call("console", "firefox", telemetry_enabled)
+                send_telemetry_ping_oop.call_args,
+                call(
+                    UsageMetrics(
```
Dexterp37, line 264: It would be great, someday, to see you use the Glean testing API and learn what you think of that. You going through the Python bindings as a first user was a very valuable experience for us!

mozregression/metrics.yaml
wlach: The list is constantly changing (e.g. `shippable` was only adding in the last year or so), so I don't think we can rely on enumeration. I could add a few more examples maybe to the list. Here's the current set mozregression supports:

```
(venv37) wlach@antwerp mozregression % mozregression --list-build-types
fennec:
  shippable
  opt
  pgo
  debug
fennec-2.3:
  shippable
  opt
  pgo
  debug
firefox:
  shippable
  opt
  pgo[linux32,linux64,win32,win64]
  debug
  asan[linux64]
  asan-debug[linux64]
gve:
  opt
  debug
jsshell:
  shippable
  opt
  pgo[linux32,linux64,win32,win64]
  debug
  asan[linux64]
  asan-debug[linux64]
thunderbird:
  opt
```

mozregression/telemetry.py
wlach: Nope!

mozregression/metrics.yaml
wlach: Expanded the list a bit while still making it clear that there are other possibilities

PR #687, opened 2020-07-12
Bug 1652339 - mozdevice pip requirement specifications should specify…

bclary:
… compatible releases not minimum ones

I think we need to do this to allow me to push mozdevice 4.0.0 to pypi without breaking current users of mozdevice which includes mozregression. I'm not sure how to update all.txt. Hopefully it will be done automatically as part of the release process.

This also updates the mozrunner requirement to be exactly 7.8.0 to prevent 7.8.1 from being picked up until we have both mozdevice 4.0.0 and mozrunner 7.8.1 published. I'll update https://github.com/mozilla/mozregression/pull/684 to reflect these changes as well.


bclary:
I'm going to close this in favor of a more minimal version.

PR #716, opened 2020-08-19
Bug 1655682 - Stop trying old taskcluster root URL

KwanEsq:
This is effectively a manual/partial backout of
f5f7466dd14e4a4e863cbc061a38025d152f7d99

Not only has the old taskcluster been decommissioned and thus no longer returns
anything, but changes to taskcluster-urls in version 13.0.1 were leading to
mozregression terminating when it passed the old TC root URLs.

### DIFF
--- a/mozregression/config.py
+++ b/mozregression/config.py
@@ -5,7 +5,6 @@ Reading and writing of the configuration file.
 from __future__ import absolute_import, print_function
 
 import os
-from datetime import datetime
 
 import mozinfo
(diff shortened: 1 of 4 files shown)


KwanEsq:
Whoops, forgot to run lint, apologies.
(Although it doesn't seem to actually catch these when I run it locally...  fortunately a new virtualenv with a pyenv 3.7 python instead of the existing system 3.6 does)

wlach:
Thanks so much!

PR #717, opened 2020-08-19
Bug 1659829 - Updates to GeckoViewExample / Fennec support

wlach:
* Remove fennec 2.3 support (the time when this was something we'd
  want to test is long past)
* Allow GVE to have shippable builds, now that we have them (fallback
  to opt if not there)
* Fix bug in fennec/gve fallback build selection

### DIFF
--- a/mozregression/fetch_configs.py
+++ b/mozregression/fetch_configs.py
@@ -485,7 +485,7 @@ class FennecIntegrationConfigMixin(IntegrationConfigMixin):
                 "opt" if build_type == "shippable" else build_type,
             )
             self._inc_used_build()
-            return
+        return
 
 
 class ThunderbirdIntegrationConfigMixin(IntegrationConfigMixin):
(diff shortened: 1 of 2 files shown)


mozregression/fetch_configs.py
```diff
@@ -485,7 +485,7 @@ def tk_routes(self, push):
                 "opt" if build_type == "shippable" else build_type,
             )
             self._inc_used_build()
-            return
+        return
```
KwanEsq, line 488: [Whoops](https://github.com/mozilla/mozregression/commit/1e599fed8543c9f37aa1f3f6865a1e37a8b370d1#diff-4ac0cc6fa4a0d1bd5c2520a4969c04c4L406-R429), thanks.

KwanEsq (approved):
:+1: 
Tested locally with `mozregression -n gve --pref "webgl.enable-surface-texture:true" -g 2020-01-01` from the bug and it successfully gets GVE onto the emulator.

PR #728, opened 2020-08-31
Bug 1661607 - Tweak the workaround to work for python 3.6 as well

theres-waldo:
In python 3.6, the %z format specifier does not match time
zone strings containing a colon, such as "+00:00".

We know the string will in practice always be "+00:00", so
just match that exactly.


wlach:
Thanks!

PR #762, opened 2020-10-26
Bug 1659568 - Only pass a copy of cmdargs to mozrunner, not the origi…

KwanEsq:
…nal object

So it can't alter it (it was adding a duplicate `--wait-for-browser` each time)


KwanEsq:
We fixed this in mozrunner instead, released in v8.0.3.

PR #797, opened 2020-12-02
Bug 1680309 - Handle invalid dates better

KwanEsq:
Don't terminate and dump a traceback.
This doesn't spot typos like 2020-003-01 that won't match the regex `re.compile(r"(\d{4})\-(\d{1,2})\-(\d{1,2})")`.  We could potentially loosen the regex, either to maybe `re.compile(r"(\d{3,5})\-(\d{1,3})\-(\d{1,3})")` or just use `\d+`.  However those at least never crashed, we just try to use them as a hash instead.

### DIFF
--- a/mozregression/dates.py
+++ b/mozregression/dates.py
@@ -8,7 +8,7 @@ import calendar
 import datetime
 import re
 
-from mozregression.errors import DateFormatError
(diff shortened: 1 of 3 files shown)


wlach (approved):
Looks great, thanks @KwanEsq!

PR #841, opened 2021-03-26
Upload release assets

wlach:
It seems like the easiest way to do this is just with CURL.


wlach:
No real way to try this other than to merge, I think.

PR #856, opened 2021-05-06
Try to figure out why glean metrics are different

wlach:


wlach:
The linter job was picking up an overly recent version of glean_parser by mistake. Won't be an issue after #855.

PR #879, opened 2021-06-24
Bug 1718065 - Handle newer GeckoView builds

wlach:


mozregression/fetch_configs.py
```diff
@@ -42,6 +42,8 @@
 # appeared on this date for m-c.
 TIMESTAMP_FENNEC_API_16 = to_utc_timestamp(datetime.datetime(2017, 8, 29, 18, 28, 36))
 
+# switch from geckoview api-16 to arm
+TIMESTAMP_GECKOVIEW_ARM = to_utc_timestamp(datetime.datetime(2021, 6, 4, 0, 0, 0))
```
wlach, line 46: This is not exactly right, but works well enough (i.e. I can download a build from 2021-06-20). Getting a more precise timestamp would be better.

agi: Copying from bugzilla:

> This is the patch that changed the name: https://hg.mozilla.org/mozilla-central/rev/b5eb71cf51d5
> 
> so it should be Sat, 05 Jun 2021 03:56:19 +0000

wlach: Thanks @agi!

PR #929, opened 2021-10-06
Allow Windows paths and `~` with --comand

mathewhodson:
This allows a Windows user to run something like this:
`mozregression -g 2021-10-04 -c ~\script.bat`


mozregression/test_runner.py
```diff
@@ -202,16 +203,25 @@ def evaluate(self, build_info, allow_back=False):
                 command = self.command.format(**variables)
             except KeyError as exc:
                 _raise_command_error(exc, " (formatting error)")
+            command = os.path.expanduser(command)
             LOG.info("Running test command: `%s`" % command)
-            cmdlist = shlex.split(command)
+            if sys.platform == "win32":
+                cmdlist = command
```
wlach, line 209: I'm not totally clear on the reason for this change. Does Windows get confused if you pass a list of expanded command line arguments?

PR #1010, opened 2022-02-15
Bug 1755112 - Merge linter requirements into main build requirements

staktrace:
The linter dependencies contained a version of click that was conflicting
with the main build dependencies, and was causing spurious warnings. This
fixes the problem.


zzzeid:
No longer needed with updated requirements workflow.

PR #1030, opened 2022-04-16
Handle initial ranges that start and end on the same day

mathewhodson:


mathewhodson:
@zzzeid Do you have a chance to review this?

zzzeid:
> @zzzeid Do you have a chance to review this?

Hi @mathewhodson -- any chance you can rebase with main and re-run the test suite? It would also be great if you can add to the description what the current behaviour is (i.e. STR) since there is no bug associated with this PR.

mathewhodson:
This fixes commands like `.\mozregression -g 20220413094328 -b 20220413215253` that would previously fail with something like this:
```
INFO: Got as far as we can go bisecting nightlies...
INFO: Last good revision: 4ddd4f1a1b2758a33b531378b2f011c3063fed1a (2022-04-13 21:52:53)
INFO: First bad revision: 4ddd4f1a1b2758a33b531378b2f011c3063fed1a (2022-04-13 21:52:53)
[...]
ERROR: The url
'https://hg.mozilla.org/mozilla-central/json-pushes?fromchange=4ddd4f1a1b2758a33b531378b2f011c3063fed1a&tochange=4ddd4f1a1b2758a33b531378b2f011c3063fed1a'
contains no pushlog. Maybe use another range ?
```

zzzeid (changes requested):
This looks good to me, minus the `TestResults` in `.gitignore`. Is this something you added to your local environment? I also am not sure we need to include `.vs` in there since this is based on personal preference of editor, though take it or leave it.

zzzeid (approved):

mathewhodson:
Thanks for the merge

PR #1043, opened 2022-05-17
requirements: add new workflow to generate requirements (bug 1767079)

zzzeid:
- add new (semi-)reusable compile-requirements workflow
- run workflow for mac/linux/windows
- upgrade some packages for compatiblity
- pin versions in base requirements
- update gitignore
- add automatically generated requirements
- fix requirements in build workflow to point to new locations
- fix build errors
- fix addons editor tests on mac


.github/workflows/compile-requirements.yml
```diff
@@ -0,0 +1,82 @@
+name: Generate requirement files
```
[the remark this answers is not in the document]
jmaher: I am not familiar with this file specifically (only loosely familiar with github workflows), so some of this is a rubber stamp

.github/workflows/compile-requirements.yml
```diff
@@ -0,0 +1,82 @@
...
+    strategy:
+      matrix:
+        os: ${{ fromJSON(inputs.os) }}
+        python-version: ${{ fromJSON(inputs.python) }}
+        requirements_file: ${{ fromJSON(inputs.requirements_files) }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+      - name: Set up Python ${{ matrix.python-version }}
```
jmaher, line 39: above it is `fromJSON(inputs.python)`, but here it is `matrix.python-version`?  I would expect these to be the same- My best guess is that the inputs.* is defining the variables for matrix.*.  If that is correct, please confirm.

.github/workflows/compile-requirements.yml
```diff
@@ -0,0 +1,82 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install --upgrade pip-tools
+      - name: Generate
+        # Generate filename based on Python version and OS.
+        run: |
+          cd ${{ inputs.requirements_directory }}
```
jmaher, line 50: why is this not `fromJSON(inputs.requirements_directory)`?

.github/workflows/run-compile-requirements.yml
```diff
@@ -0,0 +1,32 @@
...
+  pull_request:
+    branches: [requirements]
+
+jobs:
+  call-compile-requirements-linux:
+    uses: ./.github/workflows/compile-requirements.yml
+    with:
+      requirements_files: '["base.in dev.in gui-dev.in gui.in linters.in"]'
+      os: '["ubuntu-latest"]'
+      python: '["3.6", "3.7", "3.8", "3.9"]'
```
jmaher, line 16: do we need all the python versions only for linux?  I would think we could just do 3.9 only.

gui/tests/test_addons_editor.py
```diff
@@ -10,6 +11,8 @@
 
 @pytest.fixture
 def addons_editor(qtbot):
+    if os.uname().sysname.lower() == "darwin":
+        os.environ["QT_MAC_WANTS_LAYER"] = "1"
```
jmaher, line 15: I assume this will work for both intel and apple_silicon?

jmaher (approved):
overall r+, a few comments that might yield a small change or two.

.github/workflows/compile-requirements.yml
zzzeid: Yes correct -- `inputs` unfortunately can only be passed as strings (from the workflow that is calling this workflow) and we convert them to a list to be used in `matrix` above. Here were are using the value directly from the matrix run.

.github/workflows/compile-requirements.yml
zzzeid: `inputs.requirements_directory` is a string, and doesn't need to be converted back to a list or anything else.

.github/workflows/run-compile-requirements.yml
zzzeid: That's a good point. I am currently mimicking what was in the original test runs (currently all these versions of Python in Linux are tested, as well as the two other tests on Mac and Windows). We should revisit whether we need more tests (e.g. different python versions on Windows/Mac) or less tests (e.g. only one version on Linux).

Perhaps this can be informed by more telemetry (e.g. [bug 1767534](https://bugzilla.mozilla.org/show_bug.cgi?id=1767534)).

gui/tests/test_addons_editor.py
zzzeid: I'm not sure, however we currently do not have any tests for Apple silicon. I know of some bug reports on that platform.

We will need to figure out a way to add tests for Apple silicon but neither GitHub Actions nor CircleCI support that easily at this time.

PR #1061, opened 2022-07-07
workflows: use virtual environment where possible (bug 1778539)

zzzeid:
Inconsistency in the way various Python commands are being run in different workflows and jobs on different platforms can cause unexpected behaviour. For example, on macOS when running build-and-test-mac-gui, the build step fails as it can not find the mozregression package that was installed in the previous step.


dklawren (approved):
LGTM and test output looks sane. r=dkl

PR #1070, opened 2022-07-18
github: disable dependabot version updates (bug 1780012)

zzzeid:
The current dependabot version updates does not allow configuring different Python versions for different requirement files. Until this bug is fixed, dependabot version updates (non-security) should be disabled.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,0 @@
-version: 2
-updates:
-- package-ecosystem: pip
-  directory: "/"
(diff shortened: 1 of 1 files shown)


zzzeid:
Closing for now as some of the pull requests still pass (e.g. for packages that behave the same on all platforms).

zzzeid:
See also: https://github.com/dependabot/dependabot-core/issues/4606 and https://github.com/dependabot/dependabot-core/issues/4607.

PR #1072, opened 2022-07-25
tests: set qt qpa platform to offscreen for gui tests (bug 1780041)

zzzeid:


gui/tests/__init__.py
```diff
@@ -1,3 +1,9 @@
+import os
+import sys
+
 from PySide2.QtWidgets import QApplication
 
+if sys.platform != "darwin":
```
jmaher, line 6: this would be for linux && windows?  AIUI, this should only be for linux

zzzeid: hmm, good point. We don't have gui tests on Windows in CI (maybe we should?) but regardless, I'll change this to be linux-specific instead.

jmaher (approved):

PR #1073, opened 2022-07-26
requirements: add Python 3.10 in requirements (bug 1781421)

zzzeid:


.github/workflows/run-compile-requirements.yml
```diff
@@ -13,20 +13,20 @@ jobs:
     with:
       requirements_files: '["base.in dev.in gui-dev.in gui.in linters.in"]'
       os: '["ubuntu-latest"]'
-      python: '["3.6", "3.7", "3.8", "3.9"]'
+      python: '["3.6", "3.7", "3.8", "3.9", "3.10"]'
```
jmaher, line 16: can we remove 3.6 and 3.7 from this list?

jmaher (approved):
one little nit

.github/workflows/run-compile-requirements.yml
zzzeid: There's a bug tracking Python 3.6 removal (https://bugzilla.mozilla.org/show_bug.cgi?id=1767535) -- we should remove this from here after we drop support for Python 3.6 from the package (at which time we can also remove the tests/CI that run on Python 3.6).

As for Python 3.7; I would say the same as above but there is less of an urgency to drop that one since most if not all of the dependencies still support Python 3.7 as of this time.

PR #1080, opened 2022-08-08
launchers: sign code object if not signed at all (bug 1767139)

zzzeid:
Certain builds are unsigned, and when running them on arm64, macOS will
instantly terminate those processes. This patch will sign the code
object with an "ad-hoc" signature if it is unsigned.

NOTE: only unsigned packages are resigned, since signed packages with an
invalid signature should be left as-is for the time being.

### DIFF
--- a/mozregression/launchers.py
+++ b/mozregression/launchers.py
@@ -197,10 +197,12 @@ class MozRunnerLauncher(Launcher):
             return CodesignResult.PASS
         elif exit_code == 1:
             if b"code object is not signed at all" in output:
+                # NOTE: this output message was tested on macOS 11, 12, and 13.
                 return CodesignResult.UNSIGNED
             else:
                 return CodesignResult.INVALID
-        # Remaining codes normally mean the command was called with incorrect arguments.
+        # NOTE: Remaining codes normally mean the command was called with incorrect
+        # arguments or if there is any other unforeseen issue with running the command.
         return CodesignResult.OTHER
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


tests/unit/test_launchers.py
```diff
@@ -178,8 +178,126 @@ def test_launcher_deleted_whith_statement(self):
         self.assertFalse(os.path.isdir(tempdir))
 
 
-@pytest.mark.skipif(sys.platform == "darwin", reason="fails on macosx")
```
zzzeid, line 181: FYI [this change](https://github.com/mozilla/mozregression/pull/1080/files#diff-7246e1ea17f70a16641a4c92fd5483979a1bb1297111644f6e68886b19201ef5R239) was made as part of fixing the failures of this test (`test_firefox_install`).

mozregression/launchers.py
```diff
@@ -173,6 +177,39 @@ class MozRunnerLauncher(Launcher):
...
+                return CodesignResult.INVALID
+        # Remaining codes normally mean the command was called with incorrect arguments.
+        return CodesignResult.OTHER
+
+    @staticmethod
+    def _codesign_sign(appdir):
+        """Calls `codesign` to sign `appdir` with ad-hoc identity."""
+        if mozinfo.os != "mac":
+            raise Exception("_codesign_sign should only be called on macOS.")
+        return call(["codesign", "--force", "--deep", "--sign", "-", appdir])
```
jmaher, line 211: are there risks of different versions of `codesign` having different arguments and/or different output messages which could affect this?  Maybe indicate with a comment which os version and codesign version you tested with?

jmaher (approved):
great stuff!

mozregression/launchers.py
zzzeid: Good point, will add a comment.

zzzeid: I checked the documentation for all versions (going back to Mojave - https://www.unix.com/man-page/mojave/1/codesign/) and the arguments appear to all be the same. In terms of the output message, I was only able to check macOS 11, 12, and 13 which are the only ones I have access to. The output message is the same in all.

PR #1083, opened 2022-08-11
gui: upgrade to PySide6 (bug 1782252)

zzzeid:
- upgrade to PySide6
- various renamed methods related to PySide6
- fix deprecated method calls
- upgrade CI images to use ubuntu 20.04
- remove Python 3.6 support

NOTE: This pull request would target a pre-release of version `5.0.0rc0`.


.github/workflows/build.yml
```diff
@@ -13,7 +13,7 @@ on:
 
 jobs:
   lint:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
jmaher, line 16: would 22.04 be an option?

.github/workflows/build.yml
```diff
@@ -39,10 +39,14 @@ jobs:
     strategy:
       matrix:
         os:
-          - ubuntu-18.04
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+          - ubuntu-20.04
```
jmaher, line 42: same here, 22.04 maybe?

.github/workflows/build.yml
```diff
@@ -52,16 +56,16 @@ jobs:
...
+          python -m pip install -e .
       - name: Test
         run: |
           source env/bin/activate
           coverage run -m pytest tests
           coveralls --service=github
 
   build-and-test-linux-gui:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
jmaher, line 68: more 20.04 -> 22.04 possibilities

.github/workflows/build.yml
```diff
@@ -70,7 +74,8 @@ jobs:
       - uses: actions/checkout@v2
       - name: setup
         run: |
-          sudo apt install libxkbcommon-x11-0 libxcb-icccm4 libxcb-image0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-xinerama0 libxcb-xfixes0
+          sudo apt-get update
+          sudo apt install libxkbcommon-x11-0 libxcb-icccm4 libxcb-image0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-xinerama0 libxcb-xfixes0 libegl-dev -y
```
jmaher, line 78: why -dev?

jmaher (approved):

.github/workflows/build.yml
zzzeid: This fixes an import error related to Qt:
```
ImportError: libEGL.so.1: cannot open shared object file: No such file or directory
```

.github/workflows/build.yml
zzzeid: Sorry, just realized you made this comment on `lint`, and there's no reason not to use `ubuntu-latest` here. As for the one below (with the matrix), we can probably just add `ubuntu-22.04` to the matrix. GUI tests can probably use a matrix as well, maybe with a couple of versions of Python as well.

I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1785095) to track this.

jmaher (approved):
thanks!

cgsheeh (approved):
LGTM!

PR #1084, opened 2022-08-13
launchers: Disable TB update by policy (bug 1783694)

kevinoid:
Although the `app.update.auto:false` and `app.update.enabled:false` preferences are set, these have no effect on recent versions of Thunderbird due to their removal in [Bug 1420514](https://bugzilla.mozilla.org/show_bug.cgi?id=1420514) (ported to TB in [Bug 1479289](https://bugzilla.mozilla.org/show_bug.cgi?id=1479289)).  The replacement policy mechanism of setting `policies.DisableAppUpdate:true` in `distribution/policies.json` was added to `FirefoxLauncher` in 68ece936, but was not added to `ThunderbirdLauncher`.  This PR applies the same change so that app updates are disabled for Thunderbird.

Thanks for considering,
Kevin

### DIFF
--- a/mozregression/launchers.py
+++ b/mozregression/launchers.py
@@ -402,6 +402,10 @@ class ThunderbirdRegressionProfile(ThunderbirdProfile):
 class ThunderbirdLauncher(MozRunnerLauncher):
     profile_class = ThunderbirdRegressionProfile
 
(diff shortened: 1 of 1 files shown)


zzzeid (approved):
looks good to me!

PR #1092, opened 2022-08-17
gui: upgrades, enhancements, and deprecations

zzzeid:
- upgrade to PySide6 (bug 1782252)
- various renamed methods related to PySide6
- fix deprecated method calls
- upgrade CI images to use ubuntu 20.04 (bug 1786367)
- remove Python 3.6 support (bug 1767535)
- keep bundle intact to prevent corrupt signature (bug 1786365)
- move windows build + test to github actions (bug 1786362)
- move gui uploads to separate workflow (bug 1786363)


README.md
```diff
@@ -17,7 +17,6 @@ Build status:
 - Linux:
   [![Linux Build Status](https://travis-ci.org/mozilla/mozregression.svg?branch=master)](https://travis-ci.org/mozilla/mozregression)
   [![Coverage Status](https://img.shields.io/coveralls/mozilla/mozregression.svg)](https://coveralls.io/r/mozilla/mozregression)
-- Windows: [![Windows Build status](https://ci.appveyor.com/api/projects/status/bcg7t1pt2bahggdr?svg=true)](https://ci.appveyor.com/project/wlach/mozregression/branch/master)
```
jmaher, line 20: why is windows removed?

.github/workflows/deploy-gui.yml
```diff
@@ -0,0 +1,109 @@
...
+    env:
+      DISPLAY: ":99.0"
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: setup
+        run: |
+          sudo apt-get update
+          sudo apt install libxkbcommon-x11-0 libxcb-icccm4 libxcb-image0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-xinerama0 libxcb-xfixes0 libegl-dev -y
+          /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -screen 0 1920x1200x24 -ac +extension GLX
```
cgsheeh, line 19: What does this do?

cgsheeh (approved):

jmaher (approved):
all good, one question;  I also see :sheehan has a question as well.

README.md
zzzeid: So the windows build was moved from AppVeyor to GitHub Actions, thus the status can just be seen in the GitHub actions status for each release (or commit). I don't think the Linux Build Status is doing anything right now (not even pointing to the right branch, so this should probably be removed as well).

.github/workflows/deploy-gui.yml
zzzeid: Good point, this is not needed here anymore because the tests aren't running in the deploy workflow. I copied/pasted this from the test workflow when I moved the build over here. I believe it's there so that the tests can run in a sort of mock headless mode, I will remove it though.

cgsheeh (approved):

PR #1120, opened 2022-10-18
workflows: add ubuntu 22.04 to linux tests (bug 1795878)

zzzeid:
- update build-and-test-linux-gui with os matrix


jmaher (approved):
nice- way to stay on top of it all.

PR #1141, opened 2022-11-17
cli: accept version with two segments (bug 1801205)

zzzeid:


jmaher (approved):

cgsheeh (approved):
Looks like we have the same regex in `releases.py`, perhaps could make this a variable somewhere?

zzzeid:
> Looks like we have the same regex in `releases.py`, perhaps could make this a variable somewhere?

Agreed, maybe we can include that when revamping how versions are detected since that can be improved overall (e.g., shared functionality between `tag_of_release` and `tag_of_beta`, but also detecting things like `99` and `105` as release versions, since those do not match the current regex. Although arguably the latter is a usability issue and not exactly a bug. There's also repeated regex in `cli` for beta.

PR #1142, opened 2022-11-21
launchers: handle invalid codesign on macOS 13 (bug 1781111)

zzzeid:
- add check for invalid code signature if os is >= 13

### DIFF
--- a/mozregression/launchers.py
+++ b/mozregression/launchers.py
@@ -224,11 +224,11 @@ class MozRunnerLauncher(Launcher):
             remove(self.tempdir)
             raise
 
-        binarydir = os.path.dirname(self.binary)
-        appdir = os.path.normpath(os.path.join(binarydir, "..", ".."))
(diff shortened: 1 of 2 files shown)


mozregression/launchers.py
```diff
@@ -385,6 +385,17 @@ def _install(self, dest):
         super(FirefoxLauncher, self)._install(dest)
         self._disableUpdateByPolicy()
 
+        binarydir = os.path.dirname(self.binary)
+        appdir = os.path.normpath(os.path.join(binarydir, "..", ".."))
+
+        if (
+            mozinfo.os == "mac"
+            and mozinfo.os_version.version.version[0] == 13
```
zzzeid, line 393:
```suggestion
            and mozinfo.os_version.version.version[0] >= 13
```

globau (approved):
LGTM, and confirming that it fixes the issue on macOS 13.
We can address a re-singing opt-out in a follow-up issue.

PR #1176, opened 2023-01-13
requirements: upgrade PySide6 to 6.4.2 (bug 1810219)

zzzeid:
- update packages
- add some fixes needed for new version

After upgrading to PySide6 6.4.2, some instance attributes triggered
AttributeError in some of the modules (e.g., in skip_chooser.py,
and log_report.py). Those  attributes are actually class attributes
that map to various enums elsewhere, and the way they can be accessed
has changed in the latest version.


gui/mozregui/skip_chooser.py
```diff
@@ -17,9 +17,9 @@ def __init__(self, future_build_info, x=0, y=0, selectable=True):
         QGraphicsRectItem.__init__(self, x, y, self.WIDTH, self.WIDTH)
         self.future_build_info = future_build_info
         if selectable:
-            self.setFlags(self.ItemIsSelectable | self.ItemIsFocusable)
+            self.setFlags(BuildItem.ItemIsSelectable | BuildItem.ItemIsFocusable)
```
jmaher, line 20: I don't understand where `self` -> `BuildItem`, could you explain briefly?

gui/mozregui/skip_chooser.py
```diff
@@ -117,7 +117,7 @@ def build_index(self, item):
         return self.scene.build_range.future_build_infos.index(item.future_build_info)
 
     def choose_next_build(self):
-        if self.exec() == self.Accepted:
+        if self.exec() == SkipDialog.Accepted:
```
jmaher, line 120: similar here, why is `self` -> `SkipDialog` ?

gui/mozregui/skip_chooser.py
zzzeid: So in the upgraded version of PySide6, these are class attributes and are not accessible on their respective instances. Here's an example of what happens:

```
  File "/home/runner/work/mozregression/mozregression/gui/mozregui/log_report.py", line 60, in on_log_received
    cursor_to_add.movePosition(cursor_to_add.End)
AttributeError: 'PySide6.QtGui.QTextCursor' object has no attribute 'End'
```

I couldn't find anything documented for this, however these attributes are actually proxies to other class attributes (e.g., `MoveOperation.End` or `GraphicsItemFlag.ItemIsFocusable`, etc...) and do not require the instance itself.

jmaher: thanks, I wanted to make sure this was intentional, even if it was hacking around to get to this point

jmaher (approved):
lgtm

gui/mozregui/skip_chooser.py
zzzeid: In this case, this maps to: `DialogCode.Accepted` (same story as `MoveOperation` or `GraphicsItemFlag`). These are just enums.

cgsheeh (changes requested):
Please add the reasoning behind the code changes to the commit message.

zzzeid:
FYI I found [this](https://www.qt.io/blog/qt-for-python-release-6.4-is-finally-here) (under "The new Enum system") which may be very relevant to these changes.

zzzeid:
> Please add the reasoning behind the code changes to the commit message.

Done.

cgsheeh (approved):

PR #1182, opened 2023-01-19
requirements: update minimum glean-sdk version (bug 1810534)

zzzeid:
glean-sdk 52.0.1 has a universal2 wheel which we can install in the virtual environment in order to build a universal2 bundle for mozregression-gui.

See https://github.com/pypa/pip/issues/11573 for more information on why the package has to be pinned in this way in the macOS requirements.


jmaher (approved):
thanks;  bummer about the osx hack.

PR #1185, opened 2023-01-23
log: set ALLOW_COLOR to False if not sys.stdout (bug 1811930)

zzzeid:
As of PyInstaller 5.7.0, sys.stdout is changed from NullWriter to None
for GUI applications on Windows.

### DIFF
--- a/mozregression/log.py
+++ b/mozregression/log.py
@@ -13,7 +13,7 @@ from colorama import Back, Fore, Style
 from mozlog.handlers import LogLevelFilter, StreamHandler
 from mozlog.structuredlog import StructuredLogger, set_default_logger
 
-ALLOW_COLOR = sys.stdout.isatty()
+ALLOW_COLOR = sys.stdout.isatty() if sys.stdout else False
 
 
 def _format_seconds(total):


jmaher (approved):
nice

cgsheeh (changes requested):
We use `sys.stdout` in other locations in the codebase, should we check for `None` in those usages too? For example in `init_logger` in this file we pass `sys.stdout` to `StreamHandler`, and in `download_manager.py::download_progress` we use `sys.stdout.write`.

r+ once we either update those usages or confirm we don't need them. :)

zzzeid:
> We use `sys.stdout` in other locations in the codebase, should we check for `None` in those usages too? For example in `init_logger` in this file we pass `sys.stdout` to `StreamHandler`, and in `download_manager.py::download_progress` we use `sys.stdout.write`.

`init_logger` is only used in the CLI. The reason it is erroring out is because `ALLOW_COLOR` is set on import.

Same for `BuildDownloadManager` (there is a separate, `GUIBuildDownloadManager` class that does not use download_progress as you mentioned above). See download_manager.py:384 vs. build_runner.py:46.

> 
> r+ once we either update those usages or confirm we don't need them. :)

cgsheeh (approved):
Thanks for the clarification. r+

PR #1210, opened 2023-02-10
glean: update notification email (bug 1816133)

zzzeid:


jmaher (approved):
ideally this would be an alias vs an individual- if there isn't an alias, maybe it is time to create one.

zzzeid:
> ideally this would be an alias vs an individual- if there isn't an alias, maybe it is time to create one.

Agreed -- this issue is relevant to other projects as well.

PR #1218, opened 2023-02-13
telemetry: add os versions/names, and python version (bug 1767534)

zzzeid:


mozregression/telemetry.py
```diff
@@ -14,10 +17,58 @@
...
+        except (AttributeError, IndexError):
+            pass
+    elif mozinfo.os == "windows":
+        try:
+            # Fetch "version" from tuple containing Windows version information.
+            # See https://docs.python.org/3/library/platform.html#windows-platform.
+            info["windows_version"] = platform.win32_ver()[1]
+        except (AttributeError, IndexError):
+            pass
+    else:
```
cgsheeh, line 61: Nit: better to be explicit about the OS here with an `elif mozinfo.os == "linux"` instead of leaving linux in the `else` clause.

mozregression/telemetry.py
```diff
@@ -14,10 +17,58 @@
...
+        "python_version": None,
+    }
+
+    if mozinfo.os == "mac":
+        try:
+            # Fetch the "release" from tuple containing macOS version information.
+            # See https://docs.python.org/3/library/platform.html#macos-platform.
+            info["mac_version"] = platform.mac_ver()[0]
+        except (AttributeError, IndexError):
+            pass
```
cgsheeh, line 53: Question: could we pass a value like "unknown version" instead of just leaving the field blank here? It would be an annoying bug to track down if telemetry wasn't gathered for a platform and we didn't know why.

cgsheeh (changes requested):
LGTM, but have a question before r+.

PR #1219, opened 2023-02-15
launchers: handle invalid codesign on macOS 13 (bug 1817001)

zzzeid:
Move common functionality to `MozRunnerLauncher` and call it in both `FirefoxLauncher` and `ThunderbirdLauncher`.

### DIFF
--- a/mozregression/launchers.py
+++ b/mozregression/launchers.py
@@ -214,6 +214,15 @@ class MozRunnerLauncher(Launcher):
         # arguments since macOS 10, however this was tested on macOS 12.
         return call(["codesign", "--force", "--deep", "--sign", "-", appdir])
 
+    @property
+    def _codesign_invalid_on_macOS_13(self):
+        """Return True if codesign verify fails on macOS 13+, otherwise reutrn False."""
+        return (
+            mozinfo.os == "mac"
+            and mozinfo.os_version >= "13.0"
+            and self._codesign_verify(self.appdir) == CodesignResult.INVALID
+        )
+
     def _install(self, dest):
         self.tempdir = safe_mkdtemp()
         try:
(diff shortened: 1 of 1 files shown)


mozregression/launchers.py
```diff
@@ -214,6 +214,15 @@ def _codesign_sign(appdir):
         # arguments since macOS 10, however this was tested on macOS 12.
         return call(["codesign", "--force", "--deep", "--sign", "-", appdir])
 
+    @property
+    def _codesign_invalid_on_macOS_13(self):
+        """Return True if codesign verify fails on macOS 13+, otherwise reutrn False."""
```
zzzeid, line 219:
```suggestion
        """Return True if codesign verify fails on macOS 13+, otherwise return False."""
```

mozregression/launchers.py
```diff
@@ -414,6 +419,10 @@ def _install(self, dest):
         super(ThunderbirdLauncher, self)._install(dest)
         self._disableUpdateByPolicy()
 
+        if self._codesign_invalid_on_macOS_13:
+            LOG.warning(f"codesign verification failed for {self.appdir}, resigning...")
```
cgsheeh, line 423: Nit: this should probably be `re-signing`.

cgsheeh (dismissed):
LGTM.

mozregression/launchers.py
zzzeid, line 423:
```suggestion
            LOG.warning(f"codesign verification failed for {self.appdir}, re-signing...")
```

mozregression/launchers.py
```diff
@@ -385,11 +394,7 @@ def _install(self, dest):
         super(FirefoxLauncher, self)._install(dest)
         self._disableUpdateByPolicy()
 
-        if (
-            mozinfo.os == "mac"
-            and mozinfo.os_version >= "13.0"
-            and self._codesign_verify(self.appdir) == CodesignResult.INVALID
-        ):
+        if self._codesign_invalid_on_macOS_13:
             LOG.warning(f"codesign verification failed for {self.appdir}, resigning...")
```
zzzeid, line 398:
```suggestion
            LOG.warning(f"codesign verification failed for {self.appdir}, re-signing...")
```

cgsheeh (approved):

PR #1290, opened 2023-04-29
gui: improve usability in dark mode (bug 1719424)

kthchew:
Prior to this change, many elements were difficult to see in dark mode. I made various color changes if dark mode is detected to be enabled (i.e. the background color of the window is dark enough) that make them more visible.

Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1719424

Screenshot 2023-04-29 at 12 15 02 AM

I've tested my PR in both light and dark mode on macOS, ~but not on other operating systems~. edit: tested on Windows 11 and it seems to look fine (it's always in light mode there but the colors are unchanged in light mode)


gui/mozregui/mainwindow.py
```diff
@@ -75,6 +76,13 @@ def __init__(self):
         self.ui.log_view.setVisible(True)
         self.ui.logDockWidget.setVisible(True)
 
+        # Use better toolbar icons in dark mode
+        dark_mode_enabled = QGuiApplication.palette().color(QPalette.Window).lightness() < 128
```
zzzeid, line 80: This code is repeated in `gui/mozregui/report.py`, `gui/mozregui/report_delegate.py`, and `gui/mozregui/skip_chooser.py`. Perhaps it can be moved somewhere where it can be reused.

gui/mozregui/mainwindow.py
```diff
@@ -75,6 +76,13 @@ def __init__(self):
...
+        if dark_mode_enabled:
+            self.ui.actionRun_a_single_build.setIcon(QIcon(":/s_white.png"))
+            self.ui.actionStart_a_new_bisection.setIcon(QIcon(":/cutting_white.png"))
+            self.ui.actionStop_the_bisection.setIcon(QIcon(":/stop_white.png"))
```
zzzeid, line 84: To follow existing conventions, it would be better to set these icons in `gui/mozregui/mainwindow.py`, replacing the existing `icon`, `icon1`, and `icon2` values for the filename, or possibly creating "dark" versions of those variables and using those instead.

gui/mozregui/report.py
```diff
@@ -1,17 +1,24 @@
...
     "b": QColor(250, 113, 113),  # light red
     "s": QColor(253, 248, 107),  # light yellow
     "r": QColor(225, 225, 225),  # light gray
 }
+VERDICT_TO_ROW_COLORS_DARK_MODE = {
+    "g": QColor(48, 209, 88),  # green
+    "b": QColor(255, 70, 58),  # red
+    "s": QColor(160, 90, 0),  # yellow
+    "r": QColor(45, 45, 45),  # gray
+}
```
zzzeid, line 21: I'm not sure these belong in a separate dictionary. They could potentially be added with different keys to `VERDICT_TO_ROW_COLORS`.

zzzeid (changes requested):
Thanks for working on this fix -- this is definitely a big improvement over the current functionality. I noticed that the colours and icons do not transition when dark mode is toggled (unlike the colour of the window), it would be nice if the transitioning also affected these new icons and colours.

This should also be tested on Linux -- however, if there is no effect on Windows and/or Linux, this new code flow could also be applied only when macOS is detected.

Just a few minor comments otherwise, let me know if you have any questions.

gui/mozregui/mainwindow.py
kthchew: > it would be better to set these icons in `gui/mozregui/mainwindow.py`

For clarification, do you mean `gui/mozregui/ui/mainwindow.py`? That's the only place where I was able to find `icon1` etc but that seems to be auto generated. This file is already `gui/mozregui/mainwindow.py`

kthchew:
> I noticed that the colours and icons do not transition when dark mode is toggled (unlike the colour of the window), it would be nice if the transitioning also affected these new icons and colours.

I noticed that as well, but spent quite a bit of time looking at the docs trying to figure out how to detect when the theme has changed and was unable to find anything. I'll try looking again.



> however, if there is no effect on Windows and/or Linux, this new code flow could also be applied only when macOS is detected.

It may be a good idea to have the code there regardless, as if Qt [automatically detects dark mode in the future](https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5) on other operating systems it would preemptively solve the issue there too. I'll see if I can test this on Linux soon, though.

zzzeid:
> I noticed that as well, but spent quite a bit of time looking at the docs trying to figure out how to detect when the theme has changed and was unable to find anything. I'll try looking again.

I think [QEvent::PaletteChange](https://doc.qt.io/qt-6/qevent.html) maybe the right even to look for here? You would have to move the mode detection somewhere more central and trigger icon updates, etc. there on every palette change event.

> It may be a good idea to have the code there regardless, as if Qt [automatically detects dark mode in the future](https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5) on other operating systems it would preemptively solve the issue there too. ~I'll see if I can test this on Linux soon, though.~ edit: looks like dark mode is already effective on Ubuntu 22.04 (GNOME) already, and my fix is effective:

Thanks for checking, and I agree if the code works (or will potentially work in the future) let's keep the functionality.

kthchew:
> I think QEvent::PaletteChange maybe the right even to look for here?

Thanks for the pointer! I was able to figure out how to do it and have pushed my changes. I've also worked on some of your other suggested changes, as well.

gui/mozregui/mainwindow.py
zzzeid: Realized that I never replied to this.

> For clarification, do you mean `gui/mozregui/ui/mainwindow.py`? That's the only place where I was able to find `icon1` etc but that seems to be auto generated and I'm not sure I can accomplish this in the corresponding `.ui` file. This file is already `gui/mozregui/mainwindow.py`

Yes you are correct -- what I meant to say is "in the corresponding `.ui` file" of `gui/mozregui/ui/mainwindow.py`. That is probably nontrivial as you mentioned (if at all possible) since there isn't a context we can assign the different icons to in that file (even though multiple images can be assigned to the same icon).

A possible alternate approach without adding new icons that aren't in the `.ui` file could be to change the colour of the bitmap, e.g., by means of using a mask, or inverting the colours. But I think what you have right now is good as well.

zzzeid (approved):

zzzeid:
@kthchew thanks for the contribution. This will go out in the next release.

PR #1296, opened 2023-05-04
workflows: run compile jobs in parallel (bug 1810224)

zzzeid:
Previously, jobs were run in series to avoid two or more jobs pushing at the same time. This can be handled by rebasing until a push is successful while running jobs in parallel, saving a significant amount of time.

### DIFF
--- a/.github/workflows/compile-requirements.yml
+++ b/.github/workflows/compile-requirements.yml
@@ -79,4 +79,4 @@ jobs:
           find temp-requirements -name "requirements-*" -type f -exec mv -f -t requirements {} +
           git add -A
           git commit -m "Automatically generated requirements"
-          git push -f https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
+          until git push -f https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git; do git pull --rebase; done
--- a/.github/workflows/run-compile-requirements.yml
+++ b/.github/workflows/run-compile-requirements.yml
@@ -16,7 +16,6 @@ jobs:
       python: '["3.7", "3.8", "3.9", "3.10", "3.11"]'
 
   call-compile-requirements-windows:
-    needs: call-compile-requirements-linux
     uses: ./.github/workflows/compile-requirements.yml
     with:
       requirements_files: '["base.in dev.in gui-dev.in gui.in linters.in"]'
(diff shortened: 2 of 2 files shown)


jmaher:
I really don't understand the logic here.  Maybe I don't understand the problem :)

zzzeid:
> I really don't understand the logic here. Maybe I don't understand the problem :)

Previously, each requirements file was generated in a separate job, a pull / rebase was done, and then pushed back to the branch. Each job had to wait for the previous one to finish before starting. With this change, all jobs will run at the same time (i.e., compile requirements for various platforms etc.) and try to push when they are done. If they fail, they will rebase and then try to push again.

zzzeid:
In parallel: https://github.com/mozilla/mozregression/actions/runs/4883025899 (3m46s)
In series: https://github.com/mozilla/mozregression/actions/runs/4882910705 (6m49s)

.github/workflows/compile-requirements.yml
```diff
@@ -79,4 +79,4 @@ jobs:
           find temp-requirements -name "requirements-*" -type f -exec mv -f -t requirements {} +
           git add -A
           git commit -m "Automatically generated requirements"
-          git push -f https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
+          until git push -f https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git; do git pull --rebase; done
```
cgsheeh, line 82: Could we just `git pull --rebase` once and then `git push` once, or force this job to run after the compile jobs complete? I don't fully understand why we need to pull/rebase in a loop here. Is there a possibility of this running indefinitely if the pull/rebase fails for some reason?

zzzeid: There shouldn't by any possibility of running indefinitely. The push in job B will fail if job A pushes after job B rebases but before job B pushes, and there is a certain probability of this happening that is non-zero. Because of the way the jobs are self-contained (and running on different platforms), we cannot combine the commits and then push them one by one (or altogether), this is why every job has it's own "commit and push" step.

The loop is to handle the job A and job B example case I illustrated above, and in theory it could run up to two times (at least in the current setup) in one job (e.g., if two different jobs rebase and push between the rebase and the push of the third job).

zzzeid: In theory, we could create temporary branches where we push these commits to, and then combine them all in a different job, but I don't think it's necessarily worth that complexity at this point in time.

cgsheeh (approved):

PR #1319, opened 2023-05-31
gui: update __init__ to use super() (bug 1836097)

zzzeid:
See: https://bugreports.qt.io/browse/PYSIDE-2304 and https://bugreports.qt.io/browse/PYSIDE-1564. For relevant release/dev notes: https://www.qt.io/blog/qt-for-python-6.5 and https://wiki.qt.io/Qt_for_Python_Development_Notes#20._April_2023.

This only affects classes with multiple base classes.

### DIFF
--- a/gui/mozregui/bisection.py
+++ b/gui/mozregui/bisection.py
@@ -33,8 +33,12 @@ class GuiBisector(QObject, Bisector):
     handle_merge = Signal(object, str, str, str)
 
     def __init__(self, fetch_config, test_runner, download_manager, download_in_background=True):
-        QObject.__init__(self)
-        Bisector.__init__(self, fetch_config, test_runner, download_manager)
+        super().__init__(
+            fetch_config=fetch_config,
+            test_runner=test_runner,
+            download_manager=download_manager,
+            dl_in_background=download_in_background,
+        )
         self.bisection = None
         self.mid = None
         self.build_infos = None
--- a/gui/mozregui/build_runner.py
+++ b/gui/mozregui/build_runner.py
@@ -54,7 +54,7 @@ class GuiTestRunner(QObject):
     evaluate_finished = Signal()
 
     def __init__(self):
-        QObject.__init__(self)
+        super().__init__()
         self.verdict = None
         self.launcher = None
         self.launcher_kwargs = {}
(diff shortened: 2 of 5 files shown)


cgsheeh (changes requested):
Is there any difference in behaviour between between `BaseClass.__init__` and `super().__init__` for classes that are only single inheritance? I think it probably makes sense to switch over all use cases to `super().__init__` if possible, regardless of whether they use multiple inheritance or not, to keep things consistent. I was also under the impression `super().__init__` is the standard way to do call the constructor for a superclass.

zzzeid:
> Is there any difference in behaviour between between `BaseClass.__init__` and `super().__init__` for classes that are only single inheritance? I think it probably makes sense to switch over all use cases to `super().__init__` if possible, regardless of whether they use multiple inheritance or not, to keep things consistent. I was also under the impression `super().__init__` is the standard way to do call the constructor for a superclass.

This bug was affecting things that have multiple base classes that also include `QObject`. I changed all of those, we should file a separate bug to modernize the rest of the code since it is not currently broken and would be outside the scope of this bug.

zzzeid:
> Is there any difference in behaviour between between `BaseClass.__init__` and `super().__init__` for classes that are only single inheritance?

For more context, this newer PySide6 release added support for this more standard way of using `super()`. Previously this would not have worked. From the dev notes:

> Actually, multiple inheritance in PySide has nothing to do with Python, but users expect the super() function to behave the same way when used in a PySide class as in a normal Python class. We have extended the __init__ function accordingly

Nonetheless we should switch everything to use `super()` to keep things more clean. Filed [bug 1836252](https://bugzilla.mozilla.org/show_bug.cgi?id=1836252).

gui/mozregui/bisection.py
```diff
@@ -33,8 +33,12 @@ class GuiBisector(QObject, Bisector):
     handle_merge = Signal(object, str, str, str)
 
     def __init__(self, fetch_config, test_runner, download_manager, download_in_background=True):
-        QObject.__init__(self)
-        Bisector.__init__(self, fetch_config, test_runner, download_manager)
+        super().__init__(
+            fetch_config=fetch_config,
+            test_runner=test_runner,
+            download_manager=download_manager,
+            dl_in_background=download_in_background,
```
cgsheeh, line 40: `dl_in_background` isn't passed as an argument anywhere, why is it needed now?

cgsheeh (approved):

PR #1323, opened 2023-06-05
build: add Applications link + arrow to fancy dmg (bug 1836791)

zzzeid:
- add dmgbuild to requirements
- add dmg_settings.py
- update build scripts

Screenshot 2023-06-05 at 2 47 40 PM

### DIFF
--- a/gui/dmg_settings.py
+++ b/gui/dmg_settings.py
@@ -1,0 +1,154 @@
+import os.path
+import plistlib
+
+# .. Useful stuff ..............................................................
+
+application = defines.get("app", "mozregression GUI.app")  # noqa: F821
+appname = os.path.basename(application)
+
+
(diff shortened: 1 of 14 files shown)


gui/dmg_settings.py
```diff
@@ -0,0 +1,154 @@
+import os.path
```
cgsheeh, line 1: This looks like a generated file, right? We need a license header if not.

cgsheeh (approved):

gui/dmg_settings.py
zzzeid: It is not. I checked however, and it doesn't look like any files in the repo have a license header. I also checked that it is OK to only have repo-level LICENSE file and stick with that if that's the current convention in the repo.

PR #1357, opened 2023-07-05
build_range: skip TC builds that have expired artifacts (bug 1841685)

zzzeid:
- do a HEAD request to check if GVE artifacts are available from taskcluster
- raise a `BuildInfoNotFound` error when artifacts are no longer available

### DIFF
--- a/mozregression/fetch_build_info.py
+++ b/mozregression/fetch_build_info.py
@@ -13,9 +13,9 @@ import re
 from datetime import datetime
 from threading import Lock, Thread
 
+import requests
 import taskcluster
 from mozlog import get_proxy_logger
-from requests import HTTPError
 from taskcluster.exceptions import TaskclusterFailure
 
 from mozregression.build_info import IntegrationBuildInfo, NightlyBuildInfo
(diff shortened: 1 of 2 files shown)


mozregression/build_range.py
```diff
@@ -35,7 +37,21 @@ def build_info(self):
...
+                    isinstance(self._build_info, BuildInfo)
+                    and self._build_info.task_id
+                    and self._build_info.app_name == "gve"
+                ):
+                    # This is a taskcluster build, check if artifact is available.
+                    # See bug 1841685.
+
+                    # URL is an alias that redirects via a 303 status code.
+                    original_url = requests.head(self._build_info.build_url)
+                    redirect_url = original_url.headers["location"]
```
globau, line 50: To add some future-proofing it's probably worth handling the case where the URL doesn't redirect (eg. it returns a 403 directly and/or the `location` header doesn't exist)

zzzeid: This is now handled properly with `allow_redirects=True`, and is also moved to a better place (to `fetch_build_info` instead of `build_range`).

mozregression/fetch_build_info.py
```diff
@@ -154,6 +154,14 @@ def find_build_info(self, push):
             raise BuildInfoNotFound(
                 "unable to find a build url for the" " changeset %r" % changeset
             )
+
+        if self.fetch_config.app_name == "gve":
+            # Check taskcluster URL to make sure artifact is still around.
+            # build_url is an alias that redirects via a 303 status code.
+            status_code = requests.head(build_url, allow_redirects=True).status_code
+            if status_code != 200:
+                error = f"Taskcluster file {build_url} could not be fetched ({status_code})."
```
cgsheeh, line 163: Nit: we didn't really try to fetch the artifact, perhaps the error should specify we just checked if it existed. Would also be nice to indicate the value in brackets is a status code, ie `... fetched (status code: {status_code})`.

cgsheeh (dismissed):

cgsheeh (approved):

PR #1384, opened 2023-07-17
workflows: add Python 3.11, Windows CLI tests to CI

zzzeid:
- add new build-and-test-windows-cli job (bug 1844111)
- skip test that fails on windows (bug 1844111)
- add Python 3.11 to test matrix (bug 1843853)

### DIFF
--- a/tests/unit/test_launchers.py
+++ b/tests/unit/test_launchers.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import os
+import sys
 import tempfile
 import unittest
 from subprocess import CalledProcessError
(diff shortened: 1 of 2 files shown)


tests/unit/test_launchers.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import os
+import sys
```
dklawren, line 4: Why was this added here?

cgsheeh (approved):

tests/unit/test_launchers.py
zzzeid: `sys.platform` is used in lines 248 and 277.

PR #1398, opened 2023-07-25
gui: update macOS entitlements (bug 1845375)

zzzeid:
See [bug 1845375](https://bugzilla.mozilla.org/show_bug.cgi?id=1845375).

### DIFF
--- a/gui/mac/entitlements.xml
+++ b/gui/mac/entitlements.xml
@@ -4,6 +4,10 @@
      Entitlements to apply during codesigning of mozregression builds.
 -->
 <plist version="1.0">
-  <dict>
-  </dict>
+    <dict>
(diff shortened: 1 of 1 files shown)


cgsheeh (approved):
Makes sense to me. How does `codesign` know to use this file? I don't see us using the `--entitlements` flag anywhere in the repo and there aren't any references to `entitlements.xml`.

zzzeid:
> Makes sense to me. How does `codesign` know to use this file? I don't see us using the `--entitlements` flag anywhere in the repo and there aren't any references to `entitlements.xml`.

That can be found here: https://github.com/mozilla-releng/adhoc-signing/blob/main/taskcluster/adhoc_taskgraph/static_task_vars.py#L19C29-L19C29.

PR #1400, opened 2023-07-27
gui: remove unused modules from app bundle (bug 1845801)

zzzeid:
This will reduce the size of the app bundle to allow for faster unpacking.

### DIFF
--- a/gui/gui.spec
+++ b/gui/gui.spec
@@ -1,76 +1,92 @@
 # -*- mode: python -*-
 import sys
-from PyInstaller.utils.hooks import (collect_all, collect_submodules)
+from PyInstaller.utils.hooks import collect_all, collect_submodules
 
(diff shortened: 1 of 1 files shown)


gui/gui.spec
```diff
@@ -1,76 +1,92 @@
...
-                strip=False,
-                upx=True,
-                name='mozregression GUI.app',
-                icon='icons/app_icon.icns',
-                bundle_identifier=None,
-                info_plist={
-                    'NSPrincipalClass': 'NSApplication',
-                    'NSHighResolutionCapable': 'True'
-                })
+    binaries = a.binaries - TOC(
```
cgsheeh, line 33: Please add some comments describing what is going on here, or move this into a well-named function.

cgsheeh (changes requested):

PR #1402, opened 2023-08-01
gui: add splash screen hook (bug 1845573)

zzzeid:
- create new BUNDLE_WITH_TK class
- add required tcl/tk files to "lib" directory in bundle
- force resigning of bundle after it is assembled
- add splash runtime hook that shows a loading message

### DIFF
--- a/gui/splash_hook.py
+++ b/gui/splash_hook.py
@@ -6,7 +6,7 @@ import subprocess
 
 class Splash:
     # Milliseconds to pass before re-checking open files.
-    CHECK_MILLISECONDS = 3000
+    CHECK_MILLISECONDS = 500
 
     def __init__(self):
         # Must import here to avoid multiprocess forking issue with Core Foundation.
@@ -28,6 +28,20 @@ class Splash:
         self.win.after(self.CHECK_MILLISECONDS, self.check_doneness)
         self.win.mainloop()
 
+    def get_lsof(self):
+        """Return a list of files that are extracted by the bootloader."""
+        out = subprocess.check_output(["lsof", "-p", self.ppid])
+        lines = out.splitlines()
+
+        # _MEI is the prefix used in the temporary folder when extracting. See
+        # pyinstaller.org/en/stable/operating-mode.html#how-the-one-file-program-works.
+        relevant_lines = [line for line in lines if b"_MEI" in line]
+
+        # Return only the filenames.
+        parsed_lines = [line.split(b" ")[-1] for line in relevant_lines]
+        parsed_lines.sort()
+        return parsed_lines
+
     def check_doneness(self):
         """Check whether the bootloader has finished loading files."""
 
@@ -35,12 +49,11 @@ class Splash:
         # and if it has not changed between those two checks, assume extraction is
         # complete.
 
-        out = subprocess.check_output(["lsof", "-p", self.ppid])
-        if not self.last_lsof:
-            self.last_lsof = out
-        if out == self.last_lsof:
+        parsed_lsof = self.get_lsof()
+        if parsed_lsof == self.last_lsof:
             self.win.destroy()
         else:
+            self.last_lsof = parsed_lsof
             self.win.after(self.CHECK_MILLISECONDS, self.check_doneness)
 
 
--- a/gui/splash_hook.py
+++ b/gui/splash_hook.py
@@ -1,7 +1,8 @@
-from mozregression import version
 import multiprocessing as mp
 import os
 import subprocess
+
+from mozregression import version
 
 
 class Splash:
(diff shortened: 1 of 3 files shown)


gui/splash_hook.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def splash():
+    """Create a Splash instance."""
+    Splash()
+
+
+if __name__ == "__main__":
+    mp.set_start_method("fork")
+    p = mp.Process(target=splash, args=())
```
zzzeid, line 54:
```suggestion
    p = mp.Process(target=splash)
```

Remnant from a previous implementation, can be removed.

gui/splash_hook.py
```diff
@@ -0,0 +1,55 @@
...
+        )
+        self.win.eval("tk::PlaceWindow . center")
+        self.win.after(self.CHECK_MILLISECONDS, self.check_doneness)
+        self.win.mainloop()
+
+    def check_doneness(self):
+        """Check whether the bootloader has finished loading files."""
+
+        # NOTE: This is a bit hacky. The output from lsof is checked every few seconds
+        # and if it has not changed between those two checks, assume extraction is
```
zzzeid, line 35: Should probably set a max timeout here, in case something wonky causes lsof to continuously provide different output.

gui/splash_hook.py
```diff
@@ -0,0 +1,55 @@
...
+        # complete.
+
+        out = subprocess.check_output(["lsof", "-p", self.ppid])
+        if not self.last_lsof:
+            self.last_lsof = out
+        if out == self.last_lsof:
```
cgsheeh, line 41: Won't this always be true on the first run? ie when we set `self.last_lsof` in the if statement above, this condition is met.

gui/splash_hook.py
```diff
@@ -0,0 +1,55 @@
...
+            self.win.destroy()
+        else:
+            self.win.after(self.CHECK_MILLISECONDS, self.check_doneness)
+
+
+def splash():
```
cgsheeh, line 47: A comment here about why we need a wrapper would be useful.

mozregression/pyinstaller.py
```diff
@@ -0,0 +1,136 @@
...
+# ----------------------------------------------------------------------------
+# Copyright (c) 2005-2023, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+# ----------------------------------------------------------------------------
```
cgsheeh, line 16: I think this needs to be GPL-only, unless I misunderstand the licensing exceptions.

cgsheeh (changes requested):

gui/splash_hook.py
zzzeid: Oops, you're correct -- this is broken/doesn't do much other than a time delay (I refactored it recently). Will fix.

mozregression/pyinstaller.py
zzzeid: I think you are right here. I'll have to check how this can be released as both Mozilla and PyInstaller GPL within a project that uses MPL. The copied parts are fairly limited (mainly lines 96 - 104, and some of `resign_binary`. I may be able to put those in a separate file altogether and keep this file MPL. This might be a better approach.

gui/splash_hook.py
zzzeid: No longer needed, since we are now checking only the extracted folder.

gui/splash_hook.py
zzzeid: This is a remnant from another approach I was taking, and is no longer necessary, I will remove it.

cgsheeh (dismissed):
LGTM.

gui/splash_hook.py
```diff
@@ -0,0 +1,64 @@
...
+        self.win.mainloop()
+
+    def get_lsof(self):
+        """Return a list of files that are extracted by the bootloader."""
+        out = subprocess.check_output(["lsof", "-p", self.ppid])
+        lines = out.splitlines()
+
+        # _MEI is the prefix used in the temporary folder when extracting. See
+        # pyinstaller.org/en/stable/operating-mode.html#how-the-one-file-program-works.
+        relevant_lines = [line for line in lines if b"_MEI" in line]
```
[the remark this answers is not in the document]
zzzeid: `_MEI` is not a prefix of the line itself, it is a prefix of a subdirectory within a directory hierarchy (depending on where the temporary folder ends up). This could be changed to something like `/_MEI` to be more specific, or, we could parse the line to extract the path, and then look for this part in the path explicitly.

cgsheeh (approved):

PR #1424, opened 2023-08-09
build: add support for windows adhoc signing (bug 1839227)

zzzeid:
- update adhoc-sign script to support windows installer
- rename unsigned installer binary
- add workflow step to print manifest

### DIFF
--- a/bin/adhoc-sign.py
+++ b/bin/adhoc-sign.py
@@ -14,37 +14,40 @@ def main(args: argparse.Namespace):
         "macOS": "https://github.com/mozilla/mozregression"
         f"/releases/download/{release}/{filename}",
         "windows": "https://github.com/mozilla/mozregression"
-        f"/releases/download/{release}/mozregression-gui.exe",
+        f"/releases/download/{release}/mozregression-gui-unsigned.exe",
     }
 
     operating_systems = {
         "macOS": ["macapp"],
+        "windows": ["autograph_authenticode_sha2"],
     }
 
     params = {}
 
-    for os, signing_formats in operating_systems.items():
-        url = urls[os]
-        response = requests.get(url)
-        if response.status_code != 200:
-            raise ValueError(f"Could not fetch {url} ({response.status_code})")
+    os = args.os
+    signing_formats = operating_systems[os]
 
-        params[os] = {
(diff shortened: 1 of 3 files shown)


bin/adhoc-sign.py
```diff
@@ -53,6 +56,7 @@ def create_parser():
     parser = argparse.ArgumentParser(description="print ad-hoc signing manifest")
     parser.add_argument("release", nargs=1, help="signing manifest release tag")
     parser.add_argument("--bug", default="0", help="optional bug number to include")
+    parser.add_argument("--os", default="macOS", help="os (windows or macOS)")
```
cgsheeh, line 59: Nit: you could use `choices` here to ensure a valid value is passed..

bin/adhoc-sign.py
```diff
@@ -14,37 +14,40 @@ def main(args: argparse.Namespace):
...
     }
 
     params = {}
 
-    for os, signing_formats in operating_systems.items():
-        url = urls[os]
-        response = requests.get(url)
-        if response.status_code != 200:
-            raise ValueError(f"Could not fetch {url} ({response.status_code})")
+    os = args.os
```
cgsheeh, line 27: Why is the loop removed?

bin/adhoc-sign.py
```diff
@@ -14,37 +14,40 @@ def main(args: argparse.Namespace):
...
-            "product": "mozregression",
-            "reason": f"Sign application bundle for mozregression {release}.",
-            "requestor": args.requestor,
-            "sha256": hashlib.sha256(response.content).hexdigest(),
-            "signing-formats": signing_formats,
-            "signingscript-notarization": True,
-        }
+    url = urls[os]
+    response = requests.get(url)
+    if response.status_code != 200:
```
cgsheeh, line 32: Nit: this is a good use case for `raise_for_status`.

cgsheeh (approved):
LGTM.

bin/adhoc-sign.py
zzzeid: In practice, this has to be done separately for each OS in different jobs. When I first implemented this I wasn't sure if one manifest can be used or not.

cgsheeh (approved):

PR #1427, opened 2023-08-16
workflows: add missing libraries for Ubuntu builds (bug 1828122)

zzzeid:
This missing library causes a Qt platform plugin error since it does not get copied over when the app is being bundled. Upgrade Python while we're at it.

### DIFF
--- a/.github/workflows/deploy-gui.yml
+++ b/.github/workflows/deploy-gui.yml
@@ -17,20 +17,26 @@ jobs:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v3
-      - name: setup
+      - name: setup Ubuntu 20.04
+        if: matrix.os == 'ubuntu-20.04'
         run: |
           sudo apt-get update
           sudo apt install libxcb-xinerama0 libegl-dev -y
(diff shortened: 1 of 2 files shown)


dklawren (dismissed):

.github/workflows/deploy-gui.yml
```diff
@@ -17,20 +17,26 @@ jobs:
...
+      - name: setup Ubuntu 22.04
+        if: matrix.os == 'ubuntu-22.04'
+        run: |
+          sudo apt-get update
+          sudo apt install libxcb-xinerama0 libxcb-cursor0 libegl-dev qt6-qpa-plugins -y
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.9
+          python-version: "3.10"
```
cgsheeh, line 33: Is this patch bumping the Python version?

cgsheeh (approved):

.github/workflows/deploy-gui.yml
zzzeid: Yes, it makes it match with the Windows build python. They should likely all be upgraded to 3.11 soon.

cgsheeh (dismissed):

cgsheeh (approved):

PR #1453, opened 2023-09-13
launchers: pass URL arg to Android launchers (bug 1852787)

zzzeid:
- add support for URL arg to be passed to the browser
- add tests

### DIFF
--- a/mozregression/launchers.py
+++ b/mozregression/launchers.py
@@ -482,7 +482,14 @@ class AndroidLauncher(Launcher):
             self.adb.rm(self.remote_profile, recursive=True)
         LOG.debug("Pushing profile to device (%s -> %s)" % (profile.profile, self.remote_profile))
         self.adb.push(profile.profile, self.remote_profile)
-        self._launch()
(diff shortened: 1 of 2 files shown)


mozregression/launchers.py
```diff
@@ -482,7 +482,14 @@ def _start(
             self.adb.rm(self.remote_profile, recursive=True)
         LOG.debug("Pushing profile to device (%s -> %s)" % (profile.profile, self.remote_profile))
         self.adb.push(profile.profile, self.remote_profile)
-        self._launch()
+        if cmdargs and len(cmdargs) == 1 and not cmdargs[0].startswith("-"):
+            url = cmdargs[0]
+        else:
+            url = None
+        if isinstance(self, (FenixLauncher, FennecLauncher, FocusLauncher)):
+            self._launch(url)
```
cgsheeh, line 490: We should future-proof this by making it `self._launch(url=url)`, in case a positional argument is added in the future.

cgsheeh (changes requested):

cgsheeh (approved):

PR #1458, opened 2023-09-15
fetch_configs, launchers: add thunderbird-l10n support (bug 1760826)

zzzeid:
- add support for thunderbird-l10n
- limit date range to when builds became available
- add tests


cgsheeh (approved):

cgsheeh:
Nit: missing an `n` on `thunderbird-l10n` in the title.

PR #1469, opened 2023-09-22
workflows: drop Python 3.7 (bug 1854134)

zzzeid:
- remove Python 3.7 from all workflows
- remove Python 3.7 requirement files
- upgrade setuptools-scm


dklawren (approved):
lgtm

PR #1476, opened 2023-09-28
requirements: pin universal2 binaries for PyInstaller (bug 1855547)

zzzeid:
Some of the updated requirements at this time prevent PyInstaller from being able to build a universal2 binary for macOS, because pip may choose a more specific wheel when installing packages on CI. Currently, the only way to force a universal2 wheel is to specify it by filename. See https://github.com/pypa/pip/issues/11573 for more on this.


.github/workflows/run-compile-requirements.yml
```diff
@@ -11,20 +11,20 @@ jobs:
...
     with:
-      requirements_files: '["base.in dev.in gui-dev.in gui.in linters.in"]'
+      requirements_files: '["base.in base.universal2.in dev.in gui-dev.in gui.in linters.in"]'
       os: '["ubuntu-latest"]'
       python: '["3.8", "3.9", "3.10", "3.11"]'
 
   call-compile-requirements-windows:
     uses: ./.github/workflows/compile-requirements.yml
     with:
-      requirements_files: '["base.in dev.in gui-dev.in gui.in linters.in"]'
```
cgsheeh, line 21: Shouldn't the `base.universal2.in` package only be specified for macos?

cgsheeh (changes requested):

.github/workflows/run-compile-requirements.yml
zzzeid: Yes - nice catch. It has no effect on the other jobs, and is thus redundant. Will remove.

requirements/requirements-3.10-Linux.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile with Python 3.10
 # by the following command:
 #
-#    pip-compile --allow-unsafe --generate-hashes --output-file=requirements-3.10-Linux.txt base.in dev.in gui-dev.in gui.in linters.in
```
cgsheeh, line 5: I think these requirements compiles need to be re-run now, the `base.universal2.in` should no longer be included.

cgsheeh (changes requested):

cgsheeh (approved):

PR #1489, opened 2023-10-04
cli: require --arch parameter for fenix and focus (bug 1856952)

zzzeid:
Without this parameter, the user will get a misleading "unable to find build info" error.


cgsheeh (approved):
LGTM. Would it make sense to set one of the arch options as a default instead?

zzzeid:
> LGTM. Would it make sense to set one of the arch options as a default instead?

Yes I think that might be a good idea as a follow up.

PR #1585, opened 2024-01-12
fetch_configs: fix MRO in FirefoxL10nConfig (bug 1874400)

zzzeid:
This was in the correct order for `ThunderbirdL10nConfig`, however this was not the case for `FirefoxL10nConfig`. This PR fixes this problem, ensuring that the definition for the build regex is defined correctly.

Edit: also move `CommonConfig` to the right in `ThunderbirdL10nConfig`, since `ThunderbirdL10nNightlyConfigMixin` may have been overridden in that case as well.

### DIFF
--- a/mozregression/fetch_configs.py
+++ b/mozregression/fetch_configs.py
@@ -626,7 +626,7 @@ class ThunderbirdConfig(
 
 
 @REGISTRY.register("thunderbird-l10n", attr_value="thunderbird")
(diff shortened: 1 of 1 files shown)


cgsheeh (approved):
Could we split these classes up so they each have one superclass?

zzzeid:
> Could we split these classes up so they each have one superclass?

Can you clarify? There are 7 classes that extend `CommonConfig`, everything else is done via mixins.

PR #1633, opened 2024-03-06
workflows: add virus scan check (bug 1881777)

zzzeid:
Add VirusTotal scan script and job in workflow.


zzzeid:
> Makes sense. Can we somehow automate submission to https://www.microsoft.com/en-us/wdsi/filesubmission too?

I haven't looked at ways of doing this. However, since not every release will result in false positives, and even in the case of such false positives, getting the results from MS malware analysis could take a long time, and thus the value of automating this may not be worthwhile. It is mainly added to help prevent publishing of Windows executables that will not run, thus saving users from needlessly upgrading/downgrading.

PR #1690, opened 2024-05-03
workflows: add Python 3.12 support

zzzeid:
- add python 3.12 to compile requirements (bug 1894904)
- update python version in macOS run
- fix macOS version to 13 (latest version, 14, does not have tk)
- update universal2 requirements
- add python 3.12 to test matrix (bug 1851620)


.github/workflows/build.yml
```diff
@@ -119,11 +119,11 @@ jobs:
           coveralls --service=github
 
   build-and-test-mac-gui:
-    runs-on: macos-latest
+    runs-on: macos-13
```
cgsheeh, line 122: We should add a comment indicating why this is locked to macOS 13.

.github/workflows/build.yml
```diff
@@ -119,11 +119,11 @@ jobs:
...
     env:
       # We need the official Python, because the GA ones only support newer macOS versions
       # The deployment target is picked up by the Python build tools automatically
-      PYTHON_VERSION: 3.11.1
+      PYTHON_VERSION: 3.12.4
```
cgsheeh, line 126: Question: why do we only need to update the Python version for macOS?

cgsheeh (changes requested):
The bug number for the PR is missing from the title. Looks good, just a few small requests/questions.

dklawren (dismissed):

zzzeid:
> The bug number for the PR is missing from the title. Looks good, just a few small requests/questions.

The bug numbers are in the commit message, as they cover two separate bugs. Workflow changes sometimes cover multiple changes, and in the past we've omitted bug numbers from the title, and added it just in the message (e.g., https://github.com/mozilla/mozregression/pull/1384).

.github/workflows/build.yml
zzzeid: It's in the commit message ("fix macOS version to 13 (latest version, 14, does not have tk)"), but can add as a comment. In general though it is better to have a fixed version to improve consistency.

.github/workflows/build.yml
zzzeid: The workflow errors that were causing problems were related to the macOS builds, which is why there are no updates to the Windows workflow in this PR. There's a bug filed to improve version coverage in the test matrix, but this here is mainly for the builds to work.

The macos-13 GA runner ships with Python 3.12, and there's also an issue with the virtual environment on macOS. This just fixes that issue by matching the Python version, though I also just filed a bug to track the virtual env bug [separately](https://bugzilla.mozilla.org/show_bug.cgi?id=1902350) (which may help answer your question too.)

cgsheeh: Makes sense, thanks for the details. :)

cgsheeh (approved):

PR #1719, opened 2024-05-29
Bug 1899515 - Inform users of maybe missing AppArmor rules

lissyx:


mozregression/main.py
```diff
@@ -308,6 +308,89 @@ def check_mozregression_version():
...
+  exit(0);
+}
+"""
+
+    import subprocess
+    import tempfile
+
+    LOG.warning("Unprivileged user namespaces might be disabled. Checking (requires g++) ...")
+
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".cc") as source:
```
lissyx, line 375: So as mentionned in the comment, this will run only on system where it's mostly certain we are blocked, but it was a suggestion from ubuntu security team that this was the most reliable way. I'm still looking for some `unshare` call that would replicate. If i cannot find one and we decide we really dont want such a flow of building code there, then we can drop it at the expense of potentially a useless warning to some users (but most of the time the warning should be correct).

PR #1734, opened 2024-06-14
workflows: fix outdated/hardcoded requirements file (bug 1902706)

zzzeid:
Another missed change to line up with the previous updates to the test workflow.

In the future, we should somehow consolidate the common aspects of these configurations into a central place or import them from elsewhere, since changes to `deploy-gui.yml` do not get tested as part of the usual CI, and only as part of deployments.


cgsheeh (approved):
Is it possible to have `build` and `deploy` as separate steps, and `deploy` depend on `build`? That would decrease the amount of code which only runs during releases.

PR #1812, opened 2024-09-07
cli, fetch_configs, gui: use arch to construct build regex (bug 1917418)

saschanaz:


mozregression/fetch_configs.py
```diff
@@ -47,7 +47,19 @@
 TIMESTAMP_GECKOVIEW_ARM = to_utc_timestamp(datetime.datetime(2021, 6, 5, 3, 56, 19))
 
 
-def get_build_regex(name, os, bits, processor, platprefix=r".*", platsuffix="", with_ext=True):
+def infer_arch_from_bits(os, bits, processor):
```
zzzeid, line 50: `os` is not actually used in this method. Also, docstring is missing.

mozregression/cli.py
```diff
@@ -285,7 +285,7 @@ def create_parser(defaults):
             "x86_64",
         ),
         default=None,
-        help=("Force alternate build (applies to GVE, Fenix, and Focus)."),
+        help=("Force alternate build (applies to Firefox, GVE, Fenix, and Focus)."),
```
zzzeid, line 288: For completeness might be good to add `Firefox-l10n` to the list as well, since that is technically treated as a separate app here.

mozregression/cli.py
```diff
@@ -585,8 +595,18 @@ def validate(self):
...
-                self.logger.warning("--arch ignored for non Android apps.")
+            if user_defined_bits:
+                self.logger.warning(
+                    "--arch and --bits are passed together. --arch will be preferred."
+                )
+
+            if options.app not in arch_options:
+                self.logger.warning(f"--arch ignored for {options.app}.")
+                options.arch = None
+            elif options.app == "firefox" and mozinfo.os == "mac":
```
zzzeid, line 606: Should this also apply for `firefox-l10n`?

mozregression/fetch_configs.py
```diff
@@ -60,18 +72,23 @@ def get_build_regex(name, os, bits, processor, platprefix=r".*", platsuffix="",
...
+            platform = r"win32"
+        ext = r"\.zip"
     elif os == "linux":
-        if bits == 64:
+        if arch == "aarch64":
+            platform, ext = r"linux-aarch64", r"\.tar.bz2"
+        elif arch == "x86_64":
             platform, ext = r"linux-x86_64", r"\.tar.bz2"
         else:
             platform, ext = r"linux-i686", r"\.tar.bz2"
```
zzzeid, line 94: nit: might as well move `ext` out of the if/elif/else block since they all have the same value, to keep things more consistent. For visibility might even be good to move it to the top of the block.

mozregression/fetch_configs.py
```diff
@@ -609,14 +631,37 @@ def build_regex(self):
...
                 self.processor,
                 platsuffix="-asan-reporter" if "asan" in self.build_type else "",
+                arch=self.arch,
             )
             + "$"
         )
 
+    def available_bits(self):
+        """
+        Returns the available architectures for this application.
```
zzzeid, line 641:
```suggestion
        Returns the available bits for this application.
```

mozregression/fetch_configs.py
```diff
@@ -609,14 +631,37 @@ def build_regex(self):
...
+        """
+        return ["x86", "x86_64", "aarch64"]
+
 
 @REGISTRY.register("firefox-l10n", attr_value="firefox")
 class FirefoxL10nConfig(L10nMixin, FirefoxL10nNightlyConfigMixin, CommonConfig):
-    pass
+    def available_bits(self):
+        """
+        Returns the available architectures for this application.
```
zzzeid, line 656:
```suggestion
        Returns the available bits for this application.
```

zzzeid (changes requested):
Thanks for the contribution!

Few minor comments. If you'd like to test a release of this (e.g., for mozregression-gui), let me know, I can trigger a full "dev" build.

mozregression/cli.py
saschanaz:
```suggestion
            elif options.app in ("firefox", "firefox-l10n") and mozinfo.os == "mac":
```

mozregression/cli.py
saschanaz:
```suggestion
        help=("Force alternate build (applies to Firefox, Firefox-l10n, GVE, Fenix, and Focus)."),
```

saschanaz:
> Thanks for the contribution!
> 
> Few minor comments. If you'd like to test a release of this (e.g., for mozregression-gui), let me know, I can trigger a full "dev" build.

Will that be different from local gui build?

saschanaz:
(Retriggering CI)

zzzeid (approved):

PR #1813, opened 2024-09-07
pyproject: fully migrate to pyproject.toml from setup.py (bug 1917431)

saschanaz:
Did not remove setup.py as black requires one. (Followng the compatibility block in https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html)

Restricting to `setuptools>=61` as it's the first version supporting pyproject.

Restricting to `requires-python = ">= 3.8.1"` as flake8 complained about mismatching python requirement.


pyproject.toml
```diff
@@ -20,3 +20,80 @@ multi_line_output = 3
...
+  "pytest-qt==4.4.0",
+]
+linters = [
+  "flake8>=6.0.0",
+  "flake8-black",
+  "glean-parser",
+  "isort",
+  "mccabe",
+  "pyflakes",
+]
```
zzzeid, line 92: I think for the most part, this section of the code will not actually be used and adding it might make it harder to maintain these lists. E.g., the "gui" is built via pyinstaller, and distributed via GitHub in CI (based on the requirements files), and the dev/linters is also installed the same way. What do you think?

setup.py
```diff
@@ -1,61 +1,3 @@
...
-    "mozversion>=2.1.0",
-    "redo>=2.0.2",
-    "requests>=2.21.0",
-    "taskcluster>=6.0.0",
-]
-
-desc = """Regression range finder for Mozilla nightly builds"""
-long_desc = """Regression range finder for Mozilla nightly builds.
-For more information see the mozregression website:
-http://mozilla.github.io/mozregression/"""
```
zzzeid, line 29: Looks like this was dropped? It could result in more blankness on the main PyPI page.

pyproject.toml
```diff
@@ -20,3 +20,80 @@ multi_line_output = 3
...
+  "mozrunner>=8.0.2",
+  "mozversion>=2.1.0",
+  "redo>=2.0.2",
+  "requests>=2.21.0",
+  "taskcluster>=6.0.0",
+]
+requires-python = ">= 3.8.1"
+license = {text = "MPL 2.0"}
+authors = [
+  { name = "Mozilla Automation and Tools Team", email = "tools@lists.mozilla.org" }
```
zzzeid, line 52: Making it more consistent with the other similar entries.

```suggestion
  {name = "Mozilla Automation and Tools Team", email = "tools@lists.mozilla.org"}
```

pyproject.toml
```diff
@@ -20,3 +20,80 @@ multi_line_output = 3
...
+]
+classifiers=[
```
zzzeid, line 54:
```suggestion
classifiers = [
```

zzzeid (changes requested):

setup.py
saschanaz: pyproject somehow does not support this anymore and accepts readme instead 👀 (line 28)

pyproject.toml
saschanaz: That depends on what package manager is used, but given the recommended installation is to use the requirements.txt, I think we can skip this here.

saschanaz, line 92:
```suggestion
```

PR #1846, opened 2024-10-14
gui: replace usage of stdlib pipes library with shlex (bug 1914478)

jdmnd:
Python 3.13 removed the already deprecated "pipes" module from the standard library.
Some files were using the "quote" function from pipes, which prior to 3.13 was a re-export of a function with the same name in "shlex".

This change imports the shlex module directly.

https://github.com/python/cpython/blob/3.12/Lib/pipes.py#L66 https://docs.python.org/3/whatsnew/3.13.html

### DIFF
--- a/gui/build.py
+++ b/gui/build.py
@@ -7,7 +7,7 @@ See python build.py --help
 import argparse
 import glob
 import os
-import pipes
+import shlex
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


Rob--W:
I don't have review access here, but if I did I would approve this. I had similar fixes in m-c at https://bugzilla.mozilla.org/show_bug.cgi?id=1926140

When this PR is merged and a new release is made, https://bugzilla.mozilla.org/show_bug.cgi?id=1914478 can be closed.

zzzeid (approved):
Thanks for the contribution!

PR #1891, opened 2024-11-29
Allow for pulling from ESR branches.

aosmond:


aosmond:
I updated the push so that it dynamically creates an ESR branch based on the version supplied, rather than hard coding the versions. While this means someone could request an invalid ESR version, e.g. 114.0esr doesn't exist, while 115.0esr does, that doesn't seem meaningfully different from using an invalid Firefox release.

PR #2017, opened 2025-07-31
requirements: update various packages

zzzeid:
- upgrade various packages
- fix upload/download of artifacts in workflow
- pin PyInstaller to avoid bug 1930980
- unpin various dev packages


shtrom (approved):
lgtm

PR #2045, opened 2025-08-08
workflows: update ubuntu image for gui build (bug 1982049)

zzzeid:
- remove ubuntu 20.04
- add ubuntu 24.04

### DIFF
--- a/.github/workflows/deploy-gui.yml
+++ b/.github/workflows/deploy-gui.yml
@@ -10,18 +10,13 @@ jobs:
     strategy:
       matrix:
         os:
-          - ubuntu-20.04
           - ubuntu-22.04
+          - ubuntu-24.04
     env:
       DISPLAY: ":99.0"
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v4
-      - name: setup Ubuntu 20.04
-        if: matrix.os == 'ubuntu-20.04'
-        run: |
-          sudo apt-get update
-          sudo apt install libxcb-xinerama0 libegl-dev -y
       - name: setup Ubuntu 22.04
         if: matrix.os == 'ubuntu-22.04'
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/upload-gui-to-workflow.yml
```diff
@@ -11,18 +11,13 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
-      - name: setup Ubuntu 20.04
-        if: matrix.os == 'ubuntu-20.04'
-        run: |
-          sudo apt-get update
-          sudo apt install libxcb-xinerama0 libegl-dev -y
       - name: setup Ubuntu 22.04
         if: matrix.os == 'ubuntu-22.04'
         run: |
```
cgsheeh, line 28: Same comment as above.

.github/workflows/deploy-gui.yml
```diff
@@ -10,18 +10,13 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
-      - name: setup Ubuntu 20.04
-        if: matrix.os == 'ubuntu-20.04'
-        run: |
-          sudo apt-get update
-          sudo apt install libxcb-xinerama0 libegl-dev -y
       - name: setup Ubuntu 22.04
         if: matrix.os == 'ubuntu-22.04'
         run: |
```
cgsheeh, line 22: I noticed we don't have a setup step for 24.04. I'm taking the passing CI to mean it isn't required, but wanted to put that on your radar in case we're missing dependencies.

zzzeid: Yeah, I removed those intentionally after testing with/without them. See https://github.com/mozilla/mozregression/compare/8e9532c3f78de1a42fee83efd69cabd1927db674..7100cc1abdd69fbd9b0471d4d73a9edf2c416a19.

zzzeid: I'll test again with a test deployment in case the deploy step is missing something (either in that list or otherwise).

PR #2142, opened 2025-11-28
Add Homebrew installation instructions for macOS GUI

loganrosen:
Now that [a cask for `mozregression-gui` was added](https://github.com/Homebrew/homebrew-cask/pull/237781), recommend Homebrew as the primary installation method for macOS users, with direct DMG download as an alternative. Include link to brew.sh for users who need to install Homebrew first.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -17,9 +17,14 @@ For the graphical interface, we provide a simple installation process:
   follow the installation process. Some anti-virus programs falsely claim that
   mozregression is a virus, you can ignore this warning: see [bug 1647533].
 - **macOS**:
-  Download the latest **mozregression-gui.dmg** file from the
-  [github releases]. Once the file is downloaded, open it and drag
-  the "mozregression GUI" icon to "Applications".
+  - **Via Homebrew (recommended)**: Install using the following command:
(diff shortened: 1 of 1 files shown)


loganrosen:
Thanks for the review, @zzzeid!

The cask has a `livecheck` block with `strategy :github_latest`, so Homebrew's tooling automatically detects new releases. Version bumps are typically handled by community volunteers, often via automated tools like [bump-packages](https://github.com/Homebrew/actions/tree/main/bump-packages). This is how all Homebrew casks work, including VS Code, Slack, etc.

On the drag-to-Applications concern: Homebrew handles this automatically. The `app` stanza copies the app to /Applications during install, and `brew upgrade` handles updates. Users don't manually drag anything.

That said, happy to flip the order and present direct download as primary with Homebrew as an alternative for those who already use it. Let me know!

janbrasna:
One observation I have is that publishing a GH Release doesn't necessarily mean it has the signed DMG available in the assets right away. It might appear later, or never for some pre-release drafts. Is the brew/cask automation checking the actual asset existence, or that's for the volunteers to verify before bumping?

PR #2154, opened 2025-12-12
Support integration branch bisection for Fenix (Bug 1857026)

moztcampbell:
This improves support of Fenix and Focus by supporting bisection without the presence of build-info txt files on archive.mozilla.org. Instead of the text files, I use the existing JsonPushes modules to determine changeset based on build timestamp extracted from path names on archive.mozilla.org. With some small fixes to the FenixConfig, it is easy to support the Integration mixin and get proper pushlog support.

The current fenix builds use taskcluster routes called 'nightly' or 'nightly-simulation' that are roughly what we'd consider to be 'shippable' builds on desktop.

### DIFF
--- a/mozregression/fetch_build_info.py
+++ b/mozregression/fetch_build_info.py
@@ -20,7 +20,7 @@ from mozlog import get_proxy_logger
 from taskcluster.exceptions import TaskclusterFailure
 
 from mozregression.build_info import IntegrationBuildInfo, NightlyBuildInfo
-from mozregression.errors import BuildInfoNotFound, MozRegressionError
+from mozregression.errors import BuildInfoNotFound, EmptyPushlogError, MozRegressionError
 from mozregression.json_pushes import JsonPushes, Push
 from mozregression.network import retry_get, url_links
 
@@ -180,6 +180,29 @@ class ChangesetInfo:
                 return ChangesetInfo(changeset)
 
         return ChangesetInfo()
+
+    @staticmethod
+    def from_pushlog(archive_urls, fetch_config):
+        """
+        Use the json-pushes API of the source control server to lookup the
+        changeset using the build timestamp.
+        """
+        build_dt = fetch_config.get_nightly_timestamp_from_url(archive_urls.build_url)
+        branch = fetch_config.get_nightly_repo(build_dt.date())
+
+        try:
+            jpushes = JsonPushes(branch=branch)
+        except MozRegressionError:
+            LOG.info(f"Repo {branch} does not support json-pushes queries")
+            return ChangesetInfo()
+
+        try:
+            push = jpushes.push_by_timestamp(build_dt)[-1]
+        except EmptyPushlogError:
+            LOG.info(f"Unable to fetch push by timestamp for {build_dt}")
+            return ChangesetInfo()
+
+        return ChangesetInfo(push.changeset, jpushes.repo_url)
 
 
 @dataclass
--- a/mozregression/fetch_build_info.py
+++ b/mozregression/fetch_build_info.py
@@ -10,6 +10,7 @@ from __future__ import absolute_import
 
 import os
 import re
+from dataclasses import dataclass
 from datetime import datetime
 from threading import Lock, Thread
 
@@ -32,37 +33,7 @@ class InfoFetcher(object):
         self.build_regex = re.compile(fetch_config.build_regex())
         self.build_info_regex = re.compile(fetch_config.build_info_regex())
 
-    def _update_build_info_from_txt(self, build_info):
-        LOG.debug("Update build info from {}".format(build_info))
-        if "build_txt_url" in build_info:
-            build_info.update(self._fetch_txt_info(build_info["build_txt_url"]))
-
-    def _fetch_txt_info(self, url):
-        """
-        Retrieve information from a build information txt file.
-
-        Returns a dict with keys repository and changeset if information
-        is found.
-        """
-        LOG.debug("Fetching txt info from {}".format(url))
-        data = {}
-        response = retry_get(url)
-        for line in response.text.splitlines():
-            if "/rev/" in line:
-                repository, changeset = line.split("/rev/")
-                data["repository"] = repository
-                data["changeset"] = changeset
-                break
-        if not data:
-            # the txt file could be in an old format:
-            # DATE CHANGESET
-            # we can try to extract that to get the changeset at least.
-            matched = re.match(r"^\d+ (\w+)$", response.text.strip())
-            if matched:
-                data["changeset"] = matched.group(1)
-        return data
-
-    def find_build_info(self, changeset_or_date, fetch_txt_info=True):
+    def find_build_info(self, changeset_or_date):
         """
         Abstract method to retrieve build information over the internet for
         one build.
@@ -172,9 +143,57 @@ class IntegrationInfoFetcher(InfoFetcher):
         )
 
 
+@dataclass
+class ChangesetInfo:
+    """
+    Result of resolving changeset info for a nightly build.
+
+    These are generated during the fetch process and then turned into a
+    `BuildInfo` before being consumed elsewhere in the system. Some sample
+    providers are included, but the fetch config's `get_nightly_changeset`
+    can provide other behaviours.
+    """
+
+    changeset: str = None
+    repo_url: str = None
+
+    @staticmethod
+    def from_nightly_txt(archive_urls):
+        """
+        Lookup changeset info from .txt file on archive server if exists.
+        """
+        if archive_urls.build_txt_url:
+            LOG.debug("Fetching txt info from {}".format(archive_urls.build_txt_url))
+
+            response = retry_get(archive_urls.build_txt_url)
+            for line in response.text.splitlines():
+                if "/rev/" in line:
+                    repo_url, changeset = line.split("/rev/")
+                    return ChangesetInfo(changeset, repo_url)
+
+            # the txt file could be in an old format:
+            # DATE CHANGESET
+            # we can try to extract that to get the changeset at least.
+            matched = re.match(r"^\d+ (\w+)$", response.text.strip())
+            if matched:
+                changeset = matched.group(1)
+                return ChangesetInfo(changeset)
+
+        return ChangesetInfo()
+
+
+@dataclass
+class ArchiveBuildUrls:
+    """Result of scraping build folder on archive server."""
+
+    build_url: str
+    build_txt_url: str
+
+
 class NightlyInfoFetcher(InfoFetcher):
     def __init__(self, fetch_config):
         InfoFetcher.__init__(self, fetch_config)
+
         self._cache_months = {}
         self._lock = Lock()
         self._fetch_lock = Lock()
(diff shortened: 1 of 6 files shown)


mozregression/fetch_configs.py
```diff
@@ -425,23 +433,22 @@ def get_nightly_repo_regex(self, date):
 
 class FenixNightlyConfigMixin(NightlyConfigMixin):
     nightly_base_repo_name = "fenix"
-    arch_regex_bits = ""
+    has_build_info = False
+    has_json_pushes = True
 
     def _get_nightly_repo(self, date):
-        return "fenix"
+        return "mozilla-central"
```
moztcampbell: This needs to be the VCS repo, which for Fenix is now mozilla-central.

mozregression/fetch_build_info.py
```diff
@@ -212,6 +212,21 @@ def _fetch_build_info_from_url(self, url, index, lst):
...
+        """
+        build_dt = self.fetch_config.get_nightly_timestamp_from_url(infos["build_url"])
+        try:
+            branch = self.fetch_config.get_nightly_repo(date)
+            jpushes = JsonPushes(branch=branch)
+            push = jpushes.push_by_timestamp(build_dt)[-1]
+            infos["repository"] = jpushes.repo_url
+            infos["changeset"] = push.changeset
+        except Exception:
+            LOG.info(f"Unable to fetch push by timestamp for {build_dt}")
```
zzzeid: Ideally this try/except block would be a lot more specific. We can probably modify it like this:

```suggestion
        try:
            branch = self.fetch_config.get_nightly_repo(date)
        except <a more specific exception, if we know it>:
            LOG.info(f"Unable to fetch push by timestamp for {build_dt}")
        else:
            jpushes = JsonPushes(branch=branch)
            push = jpushes.push_by_timestamp(build_dt)[-1]
            infos["repository"] = jpushes.repo_url
            infos["changeset"] = push.changeset
```

But I think at least we can probably move some of the items to the else block.

mozregression/fetch_configs.py
zzzeid: Will this impact historical builds, before the move to mozilla-central? We may have to add some fallback mechanisms.

mozregression/fetch_configs.py
```diff
@@ -425,23 +433,22 @@ def get_nightly_repo_regex(self, date):
...
 
     def get_nightly_repo_regex(self, date):
-        repo = self.get_nightly_repo(date)
-        repo += self.arch_regex_bits  # e.g., ".*arm64.*".
+        repo = f"{self.nightly_base_repo_name}-[^-]+-android"
+        if self.arch:
+            repo += f"-{self.arch}"
```
zzzeid, line 492: Same question for these lines, does this need to accommodate different repo regexes based on the date?

mozregression/fetch_build_info.py
```diff
@@ -277,6 +292,8 @@ def find_build_info(self, date, fetch_txt_info=True, max_workers=2):
                 infos = sorted(valid_builds, key=lambda b: b[0])[0][1]
                 if fetch_txt_info:
                     self._update_build_info_from_txt(infos)
+                if ("changeset" not in infos) and self.fetch_config.has_json_pushes:
+                    self._update_build_info_from_json_pushes(date, infos)
```
zzzeid: I think this is getting shoe-horned into the NightlyInfoFetcher when it should probably fall back to IntegrationInfoFetcher (e.g., via IntegrationHandler). I.e., we need a FenixIntegrationConfigMixin in fetch_configs.

zzzeid (changes requested):

mozregression/fetch_configs.py
moztcampbell: Old Code:
 - `nightly_base_repo_name` used to identify taskcluster routes and archive.m.o paths
 - `get_nightly_repo_regex` used `get_nightly_repo` incorrectly I'd argue
 - The `get_nightly_repo` used for desktop builds integration branches
 - Fenix doesn't support integration branch bisection

New Code:
 - `get_nightly_repo_regex` now uses `nightly_base_repo_name` since that still says "fenix" and still is used for archive.m.o paths
 - Both fenix and desktop use `get_nightly_repo` to find the source branch

So, the fenix artifacts from `fenix` repo to `mozilla-android` repo to `mozilla-central` repo all published to same taskcluster routes and same naming scheme for archive.m.o. Good question though since this part drove me a little mad to figure out what is breaking for me.

mozregression/fetch_configs.py
moztcampbell: Similarly, the scheme for archive.m.o was consistent across all repo migrations. My "fix" is to use 'nightly_base_repo_name' (which is the taskcluster/archive.m.o name) instead of 'get_nightly_repo' which is the active taskcluster repo.

mozregression/fetch_configs.py
moztcampbell: Old fenix artifacts will still be fetchable from pre-monorepo with this code since they come from archive.m.o. Any integration branch artifacts are long gone from task cluster. We never published any integration artifacts on archive.m.o as we do for Desktop (eg https://archive.mozilla.org/pub/firefox/integration/autoland/2024/12/06/09af335a2b3b0e2ccf214720645de29b51407509/) so Fenix can't do fine-grained bisection beyond the time range of taskcluster artifacts (like 3 months or so).

moztcampbell: I'll add the other repo references and comments though

moztcampbell:
I think you are right that I have misplaced the patch that has `FenixIntegrationConfigMixin`....

moztcampbell:
In the second patch in stack, I introduce `FetchInfo` types as helpers to the NightlyInfoFetcher and allow the fetch_configs to control which variant to use. This clarifies the fetch_configs that don't use an `changeset` resolution for nightlies, and the later patches in stack introduce the pushlog-based version.
I still maintain the call to json-pushes distinct from the IntegrationHandler since resolving in NightlyHandler means I get push range urls while bisecting the nightlies before the IntegrationHandler actually kicks off.
Feedback welcome on the fetch_configs interfaces. I have them return a class to control the FetchInfo type, but maybe there are better approaches.

moztcampbell:
Here is my complete stack with more testing and more bug fixes.

- I've broken this down into a stack of patches that each pass tests
- This has the FetchInfo/TxtFetchInfo/PushlogFetchInfo refactor so that a fetch config can specify which to use
- This added a few more tests for the NightlyInfoFetcher for the fenix case
- I changed how mixins are inherited to be at start of list since I was hitting MRO issues trying to override methods and these seems consistent with Python recommendations as best I can tell
- The FenixIntegrationConfig is now included ends up being simplified
- I've added more internal tests and done more manual testing of this version
- I ensured that even without integration-bisection that the nightly bisection is able to generate the correct pushlogs

mozregression/fetch_build_info.py
```diff
@@ -172,9 +142,88 @@ def find_build_info(self, push):
         )
 
 
+class FetchInfo:
+    """
+    Holds data about a candidate builds fetched from the nightly archive
+    servers.
```
zzzeid: Summary should fit on the first line and respect max width (see [pep-0257](https://peps.python.org/pep-0257/#multi-line-docstrings) for more detail). More elaborate detail can go after the blank line, if needed.

This applies to any other docstrings in this PR that are similar.

mozregression/fetch_build_info.py
```diff
@@ -172,9 +142,88 @@ def find_build_info(self, push):
...
+    def __eq__(self, other):
+        if not isinstance(other, FetchInfo):
+            return False
+        return (
+            self.build_url == other.build_url
+            and self.build_txt_url == other.build_txt_url
+            and self.changeset == other.changeset
+            and self.repository == other.repository
+        )
+
```
zzzeid: It looks like this is added just for usage in tests. It would be better to remove this and instead either add a method that is used in tests, to do what this is doing, or to simply check these things explicitly in the tests.

mozregression/fetch_build_info.py
```diff
@@ -172,9 +142,88 @@ def find_build_info(self, push):
...
+    def update_metadata(self, fetch_config):
```
zzzeid: nit:
```suggestion
    def update_info(self, fetch_config):
```

mozregression/fetch_build_info.py
```diff
@@ -183,34 +232,41 @@ def _fetch_build_info_from_url(self, url, index, lst):
...
+                    build_urls.append(link)
+                if self.build_info_regex.match(name):
+                    build_txt_urls.append(link)
+
+        if build_urls:
+            build = build_urls[0]
+            build_txt = build_txt_urls[0] if build_txt_urls else None
+
+            fetch_info_class = self.fetch_config.get_nightly_fetch_info_class()
+            fetch_info = fetch_info_class(build, build_txt)
```
zzzeid:
```suggestion
            fetch_info = self.fetch_config.fetch_info(build, build_txt)
```

mozregression/fetch_configs.py
```diff
@@ -262,6 +263,13 @@ def extra_persist_part(self):
         """
         return ""
 
+    def get_nightly_fetch_info_class(self):
```
zzzeid:
```suggestion
    @property
    def info_class(self):
```

mozregression/fetch_configs.py
```diff
@@ -424,23 +435,50 @@ def get_nightly_repo_regex(self, date):
...
+        #       repo Fenix source code was in.
+        product = self.nightly_base_repo_name
+        version = r"[^-]+"
+        repo = f"{product}-{version}-android"
+        if self.arch:
+            repo += f"-{self.arch}"
         return self._get_nightly_repo_regex(date, repo)
 
+    def get_nightly_fetch_info_class(self):
+        return PushlogFetchInfo
```
zzzeid: I think this would improve readability

```suggestion
    @property
    def info_class(self):
        return PushlogFetchInfo
```

mozregression/fetch_configs.py
```diff
@@ -684,7 +735,7 @@ class ThunderbirdL10nConfig(L10nMixin, ThunderbirdL10nNightlyConfigMixin, Common
 
 
 @REGISTRY.register("fennec")
-class FennecConfig(CommonConfig, FennecNightlyConfigMixin, FennecIntegrationConfigMixin):
+class FennecConfig(FennecNightlyConfigMixin, FennecIntegrationConfigMixin, CommonConfig):
```
zzzeid: I think this change here points to a slight problem in the implementation. Ideally we wouldn't need to "override" the `get_nightly_fetch_info_class` method. Instead, it should be provided by the correct mixin as needed. So if we want to keep this intended behaviour, I would detach `get_nightly_fetch_info_class` altogether from `CommonConfig` and it to the corresponding/necessary mixins instead. That would eliminate the need to change the order of classes here (and elsewhere.)

mozregression/json_pushes.py
```diff
@@ -155,3 +155,23 @@ def push(self, changeset, **kwargs):
...
+
+    def push_by_timestamp(self, timestamp):
+        """
+        Returns a list of Push objects for a timestamp.
+
+        This will return at least one Push. In case of error it will raise
+        a MozRegressionError.
+        """
+
+        assert isinstance(timestamp, datetime.datetime)
```
zzzeid: Use of `assert` should be avoided outside of tests.
```suggestion
        if not isinstance(timestamp, datetime.datetime):
            raise TypeError()
```

mozregression/fetch_configs.py
```diff
@@ -130,7 +131,7 @@ def __init__(self, os, bits, processor, arch):
         self.processor = processor
         self.set_arch(arch)
         self.repo = None
-        self.set_build_type("opt")
+        self.set_build_type(self.BUILD_TYPES[0])
```
zzzeid, line 134: Nice. I wonder why this wasn't set like this originally.

mozregression/fetch_build_info.py
```diff
@@ -172,9 +142,88 @@ def find_build_info(self, push):
         )
 
 
+class FetchInfo:
```
zzzeid: This is a nit, and I don't have a better suggestion off the top of my head, but I think the naming here might get a little confusing since we now have `InfoFetcher` and `FetchInfo`, and the main point of having this class is to have broader sources to fetch from. So maybe we should figure out a better separation here between the source of the info, and the info itself.

zzzeid (changes requested):

moztcampbell:
I did another pass at this and replaced `FetchInfo` I was adding with `ArchiveBuildUrls` and `ChangesetInfo` to reflect the two stages they are generated. This also made it easier to avoid passing a class around and instead using a mixin on the fetch config to control how the ChangesetInfo is computed.
In the process of doing this I discovered a new bug in the thunderbird-l10n builds where they weren't finding the text files even though they were partially configured for them.

PR #2191, opened 2026-02-09
releases: sort output of `--list-releases` (Bug 2015892)

ErichDonGubler:
`--list-releases` doesn't have a specified order for its output. It seems like this misses an opportunity to help a user find a release they actually want; I've never seen a useful version index without some form of sorting. Apply sorting by reverse order, under the assumption that the user will likely be most interested in the most recent releases.

### DIFF
--- a/mozregression/releases.py
+++ b/mozregression/releases.py
@@ -136,7 +136,9 @@ def formatted_valid_release_dates():
     the valid release dates.
     """
     message = "Valid releases: \n"
-    for key, value in releases().items():
+    releases_descending = list(releases().items())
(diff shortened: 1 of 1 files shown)


zzzeid (changes requested):
lgtm, just needs a reference to a bug in the title.

ErichDonGubler:
@zzzeid: Done! I wasn't sure if you wanted the title changed, but I did both it and the commit message.

zzzeid (approved):

zzzeid:
> @zzzeid: Done! I wasn't sure if you wanted the title changed, but I did both it and the commit message.

Thanks. There are some CI issues but will merge this after fixing the issues.

PR #2192, opened 2026-02-10
launchers: Handle existing distribution directory in Firefox builds (bug 2015791)

ahochheiden:
See also: https://bugzilla.mozilla.org/show_bug.cgi?id=2010200


zzzeid (approved):
lgtm!

PR #2197, opened 2026-06-29
test_runner: add --prompt mode driving the Firefox DevTools MCP

msujaws:
Add an agent-driven verdict mode, the natural-language equivalent of --command (like `git bisect run`). `--prompt "<instruction>"` shells out to the `claude` CLI in headless mode, pointed at the @mozilla/firefox-devtools-mcp server via a generated --mcp-config, to inspect each build and decide good/bad.

The MCP launches the build itself, so AgentTestRunner installs the build to obtain the binary path (passed via --firefox-path) without starting it. Verdicts are parsed as GOOD/BAD from the agent output.

Gating: the Firefox DevTools MCP supports Firefox 100+, so ranges that predate it are rejected. This happens both up front (resolved good/bad range in cli.validate) and per build (mozversion application_version), configurable via --prompt-min-version (default 100).

A pre-run check (Application.check_prerequisites) fails fast before bisecting if `claude`/`npx` are missing or if the instruction is not usable for a good/bad determination.

--prompt is mutually exclusive with --command and --launch. Adds --prompt-headless and --prompt-model. New UnsupportedVersionError.

### DIFF
--- a/mozregression/cli.py
+++ b/mozregression/cli.py
@@ -21,7 +21,12 @@ from mozregression import __version__
 from mozregression.branches import get_name
 from mozregression.config import DEFAULT_CONF_FNAME, get_config, write_config
 from mozregression.dates import is_date_or_datetime, parse_date, to_datetime
-from mozregression.errors import DateFormatError, MozRegressionError, UnavailableRelease
+from mozregression.errors import (
+    DateFormatError,
+    MozRegressionError,
+    UnavailableRelease,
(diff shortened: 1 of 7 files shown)


msujaws:
A sample command that can be used to demonstrate this is:
```
mozregression --prompt "Go to about:preferences and look for an AI controls menu item. Good if it is there, bad if it isn't." --find-fix
```

PR #2198, opened 2026-07-13
branches: add esr branch support (bug 2048638)

Magnolia-Liu:
This change makes the CLI + GUI recognize the mozilla-esr<N> / comm-esr<n> / esr<N> pattern and construct the hg repo URL dynamically.

Resolution is a blind HTTP fetch: we build the URL from whatever version the user passes and request it. If that ESR exists on hg the fetch is a hit and bisection proceeds. If it doesn't, the request simply misses and the branch is treated as unavailable.

Previously, mozregression was not working as expected since it could not find any builds when using commands like `mozregression --app thunderbird --repo comm-release` for Thunderbird branches other than comm-central. This was because comm-esr, comm-release, and comm-beta don't have opt builds, they only have shippable ones. 

Fix the behaviour for fetching integration builds both for comm-esr branches and other comm branches by implementing similar behaviour as existing tk_routes construction in Firefox.

Add small help text that shows how to fetch ESR branches as they will not show up in the drop down menu.

New UI:
Screenshot 2026-07-16 at 5 35 46 PM


mozregression/branches.py
zzzeid, line 89: We need to handle setting the `category` value for esr branches (it should be "releases") so that we can access integration (taskcluster) builds during bisection (or when launching based on a date or commit). Otherwise this will default to only nightly builds (which do not exist).

mozregression/branches.py
```diff
@@ -38,12 +39,21 @@ def set_alias(self, alias, branch_name):
         self._aliases[alias] = branch_name
 
     def get_url(self, branch_name_or_alias):
+        name = self.get_name(branch_name_or_alias)
         try:
-            return self._branches[self.get_name(branch_name_or_alias)]
+            return self._branches[name]
         except KeyError:
+            if RE_ESR.match(name):
+                return self.DEFAULT_REPO_URL + "releases/mozilla-esr%s" % RE_ESR.match(name).group(
```
zzzeid: We also have comm-esr* branches, so assuming `esrN` is referring to a firefox esr* may not always be desired (e.g., if `app=thunderbird` is passed).

EDIT: fixed names.

zzzeid (changes requested):

janbrasna:
FYI mentioning previous art in #1891 or f37208c in case there's any inspiration. (E.g. the `tag_of_esr()` bit there might be interesting?)

mozregression/fetch_configs.py
```diff
@@ -238,7 +238,19 @@ def set_repo(self, repo):
...
-        self.repo = branches.get_name(repo) if repo else None
+        if repo:
+            name = branches.get_name(repo)
+            if self.app_name == "thunderbird" and name.startswith("esr"):
+                self.repo = "comm-" + name
+            elif self.app_name == "firefox" and name.startswith("esr"):
+                self.repo = "mozilla-" + name
+            elif name.startswith("esr"):
+                # A bare `esr` defaults to the Firefox (`mozilla`) repository.
+                self.repo = "mozilla-" + name
```
zzzeid: nit: name.startswith("esr") is repeated 3 times, and can be determined at the beginning

mozregression/fetch_configs.py
```diff
@@ -559,11 +571,17 @@ class ThunderbirdIntegrationConfigMixin(IntegrationConfigMixin):
...
+                (
+                    ".shippable"
+                    if self.integration_branch != "comm-central"
+                    or (self.integration_branch == "comm-central" and build_type == "shippable")
+                    else ""
+                ),
                 push.changeset,
                 _common_tk_part(self),
-                build_type,
+                "opt",
```
zzzeid, line 583: Can you add a comment about this change in the PR description / why it's needed?

`self.integration_branch == "comm-central" and build_type == "shippable"` would affect existing behaviour.

mozregression/fetch_configs.py
```diff
@@ -675,7 +693,18 @@ def available_archs(self):
...
+        "opt",
+    )
+    BUILD_TYPE_FALLBACKS = {
+        "shippable": ("opt",),
+        "opt": ("shippable",),
+    }
+
+    def __init__(self, os, bits, processor, arch):
+        super(ThunderbirdConfig, self).__init__(os, bits, processor, arch)
+        pass
```
zzzeid:
```suggestion
```

CLAUDE.md
zzzeid, line 1: I think this was added by mistake?

mozregression/fetch_configs.py
```diff
@@ -675,7 +693,18 @@ def available_archs(self):
...
 ):
-    pass
+    BUILD_TYPES = (
+        "shippable",
+        "opt",
+    )
+    BUILD_TYPE_FALLBACKS = {
+        "shippable": ("opt",),
+        "opt": ("shippable",),
+    }
```
zzzeid, line 702: Same here, this would affect existing behaviour.

zzzeid (changes requested):

mozregression/fetch_configs.py
Magnolia-Liu: When I was testing the tool I noticed that there doesn't seem to be opt builds for `comm-release`, `comm-beta`, and the `comm-esr` branches. Running commands like `mozregression --app thunderbird --repo comm-release` would not find any builds unless I added .shippable in the tk_routes like how they are constructed in equivalent Firefox method. There seemed to be both opt and shippable builds for `comm-central` so I tried to add support for both while keeping opt as the default build-type for the CLI. But if that changes the existing behaviour too much I can revert to only allowing opt builds for `comm-central`.

Magnolia-Liu: But maybe I could have just been using the wrong commands for those branches?

mozregression/fetch_configs.py
Magnolia-Liu: This relates to my other reply where there doesn't seem to be opt builds for thunderbird branches other than comm-central, so I wanted to make shippable a select-able build type in the gui. Now that I think about it more though, I should probably disable the opt option for the majority of the thunderbird branches unless I'm mistaken with the build-type issue.

mozregression/fetch_configs.py
zzzeid: So this would fix the behaviour for fetching integration builds both for comm-esr branches and other comm branches that were not working as expected. Would be good to mention that in the commit message! (i.e., implementing similar behaviour as existing tk_routes for firefox)

mozregression/branches.py
```diff
@@ -38,19 +39,33 @@ def set_alias(self, alias, branch_name):
         self._aliases[alias] = branch_name
 
     def get_url(self, branch_name_or_alias):
+        name = self.get_name(branch_name_or_alias)
         try:
-            return self._branches[self.get_name(branch_name_or_alias)]
+            return self._branches[name]
         except KeyError:
+            if RE_ESR.match(name):
```
zzzeid: This check is repeated a few times in this class. We could put this functionality in `self._esr_match` and `self.is_esr_name`.

mozregression/branches.py
```diff
@@ -38,19 +39,33 @@ def set_alias(self, alias, branch_name):
...
+                url = self.DEFAULT_REPO_URL + "releases/%s" % name
+                return url
             raise MozRegressionError("No such branch '%s'." % branch_name_or_alias)
 
     def get_name(self, branch_name_or_alias):
+        if branch_name_or_alias:
+            match = RE_ESR.match(branch_name_or_alias)
```
zzzeid: nit
```suggestion
        if branch_name_or_alias and match := RE_ESR.match(branch_name_or_alias):
```

Then later lines can be dedented.

mozregression/fetch_configs.py
```diff
@@ -238,7 +238,20 @@ def set_repo(self, repo):
...
+        if repo:
+            name = branches.get_name(repo)
+            is_esr = name.startswith("esr")
+            if self.app_name == "thunderbird" and is_esr:
+                self.repo = "comm-" + name
+            elif self.app_name == "firefox" and is_esr:
+                self.repo = "mozilla-" + name
+            elif is_esr:
+                # A bare `esr` defaults to the Firefox (`mozilla`) repository.
+                self.repo = "mozilla-" + name
```
zzzeid: nit - I don't think adding the check for `self.app_name == "firefox"` is doing anything here.
```suggestion
            elif is_esr:
                # A bare `esr` defaults to the Firefox (`mozilla`) repository.
                self.repo = "mozilla-" + name
```

mozregression/fetch_configs.py
```diff
@@ -238,7 +238,20 @@ def set_repo(self, repo):
         If not set or set to None, default repos would be used (see
         :meth:`get_nightly_repo` and :attr:`integration_branch`)
         """
-        self.repo = branches.get_name(repo) if repo else None
+        if repo:
+            name = branches.get_name(repo)
+            is_esr = name.startswith("esr")
```
zzzeid, line 243: I think it would be better to reverse the logic here. Handle the general case first, and then handle the esr case.

zzzeid (changes requested):

zzzeid (approved):

